PR #11, opened 2020-06-28
Simple improvements to code and README.md

evaporei:
Some very dummy improvements to the code and `README.md`.

By the way the code on `rational_to_double` I can change to a `if/else` if you prefer that, or something different, I just think it didn't need two indentations, using `String`, `collect`, and it could be safer for free

### DIFF
--- a/src/edn/mod.rs
+++ b/src/edn/mod.rs
@@ -143,7 +143,7 @@ impl Edn {
             Edn::Int(i) => to_double(i).ok(),
             Edn::UInt(u) => to_double(u).ok(),
             Edn::Double(d) => Some(d.to_owned()),
-            Edn::Rational(r) => rational_to_double(r.to_owned()),
+            Edn::Rational(r) => rational_to_double(&r),
             Edn::Bool(_) => None,
             Edn::Nil => None,
         }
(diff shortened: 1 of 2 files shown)


src/edn/mod.rs
```diff
@@ -257,4 +259,5 @@ fn parses_rationals() {
     assert_eq!(rational_to_double(String::from("15/4")).unwrap(), 3.75f64);
     assert_eq!(rational_to_double(String::from("3 4")), None);
     assert_eq!(rational_to_double(String::from("3/4/5")), None);
-}
\ No newline at end of file
+    assert_eq!(rational_to_double(String::from("text/moretext")), None);
```
evaporei: This would panic before

src/edn/mod.rs
```diff
@@ -241,13 +241,15 @@ fn to_double<T>(i: T) -> Result<f64,std::num::ParseFloatError>
...
-        _ => None
+    if r.split('/').count() == 2 {
+        let vals = r.split('/')
+            .map(ToString::to_string)
+            .map(|v| v.parse::<f64>())
+            .map(Result::ok)
+            .collect::<Option<Vec<f64>>>()?;
+        return Some(vals[0] / vals[1]);
     }
+    None
```
evaporei, line 252: We can add a `else` if you prefer here, and remove the `return` on `Some`

naomijub (approved):
LGTM

PR #12, opened 2020-07-09
Initial steps to config Edn::Char

naomijub:
Problem:
`The EDN char format is \c for type char 'c', but Rust doesn't allow this pattern in macro_rules`

Relevant discussion to the topic https://internals.rust-lang.org/t/idea-escaping-macro-separators/7877/11


naomijub:
Chars already implemented

PR #17, opened 2020-07-10
edn: use proper Rust Set type/collection internally for Edn::Set

### LINKED ISSUE #3: Use `HashSet` for `Edn::Set`
the `set!` macro is already implemented, however initial implementation uses `Vec<Edn>`. The ideia is to convert `Edn::Set` inner type `Set` from `Vec<Edn>` to `HashSet<Edn>`

evaporei:
To actually use a `Set` Rust type internally for `Edn::Set` I had to do several changes, they are:

- I tried using `HashSet`, however since both `HashMap` and `HashSet` don't implement `Hash`, they can't be used as keys on `HashMap`s and/or values on `HashSet`s. Since this ain't' viable, I went with the `BTreeMap`/`BTreeSet` solution, which is the same as the [mentat/edn](https://github.com/mozilla/mentat/tree/master/edn) crate;
- Also since `HashSet<Edn::Double>` exists now, `Edn::Double` suddenly had to implement `Hash`. To make this happen I've used the `ordered_float` crate, just like the [mentat/edn](https://github.com/mozilla/mentat/tree/master/edn) did;
- Also I had to do some `serialization` and `macro` internal changes, because of the new types that I am using.

## Considerations on API Design

There is one thing I would like a second opinion. Before to use the variant `Edn::Double`, you could just pass a `f64` to it, like `Edn::Double(1.2)`.
Now since we will be using the `OrderedFloat`, I've created a public alias `Double` which just stands for `OrderedFloat<f64>`.

The problem with this is that the user has to either do `Edn::Double(Double(1.2))` or it can do `Edn::Double(1.2.into())` (more pleasant in my opinion). Do you think this is a problem/issue? If you are okay with the user having to do this conversion or use the new alias type then I think we can proceed normally.

By the way this is probably a **breaking change**, since `Edn::Double(1.2)` won't work anymore, and will have to be substituted by `Edn::Double(1.2.into())`. But still, is a very simple fix, and it doesn't affect macros at all, just the creation of `Edn` manually via enum 🙂 

Also, any suggestion is always welcome 😊 , the project is yours after all hahaha

Fixes https://github.com/naomijub/edn-rs/issues/3

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ fn main() {
         List::new(
             vec![
                 Edn::Symbol("sym".to_string()),
-                Edn::Double(1.2),
+                Edn::Double(1.2.into()),
                 Edn::Int(3),
                 Edn::Bool(false),
                 Edn::Key("f".to_string()),
@@ -44,7 +44,7 @@ To navigate through `Edn` data you can just use `get` and `get_mut`:
 let edn = edn!([ 1 1.2 3 {false :f nil 3/4}]);
 
 assert_eq!(edn[1], edn!(1.2));
-assert_eq!(edn[1], Edn::Double(1.2f64));
+assert_eq!(edn[1], Edn::Double(1.2f64.into()));
 assert_eq!(edn[3]["false"], edn!(:f));
 assert_eq!(edn[3]["false"], Edn::Key("f".to_string()));
 ```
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -53,15 +53,15 @@ assert_eq!(edn[3]["false"], Edn::Key("f".to_string()));
  #![recursion_limit="512"]
  #[macro_use] extern crate edn_rs;
  
- use std::collections::{HashMap, HashSet};
+ use std::collections::{BTreeMap, BTreeSet};
  use crate::edn_rs::serialize::Serialize;
  
  fn main() {
      ser_struct!{
```
naomijub, line 61: Guess this could be kept as HashMaps and HashSets as they depend only on the serialization module.
Maybe add an example with BTreeMap and BTreeSet

README.md
```diff
@@ -23,7 +23,7 @@ fn main() {
         List::new(
             vec![
                 Edn::Symbol("sym".to_string()),
-                Edn::Double(1.2),
+                Edn::Double(1.2.into()),
```
naomijub, line 27: I think `into` is better than `Double(1.2)`. Also `Double` could be set as private so people don't use this option and don't know the inner logic.

evaporei: We can do this, I only left this public because of the other variants like `Set`, `Map`, etc, so I thought it could be intuitive for the user to just do the same for Doubles, since just adding a f64 won't work.

README.md
evaporei: I will put one field for each, so 4 of them, Hash{Map, Set} and BTree{Map, Set}

README.md
naomijub: In the future we could make a PR to try this with with destructured tuples `(sign, mantissa, exponent)`

naomijub: Oh, OK.
We need to filter what is public and what is private in this crate. For now almost everything is public kkkkk

README.md
evaporei: Solved, by you!

README.md
evaporei: For now, we will let it public, then later we will filtrate what is not necessary.

PR #18, opened 2020-07-15
Parses Str into EDN type

naomijub:


evaporei (changes requested):
No

evaporei (approved):
JK, I've coded with you

PR #20, opened 2020-07-17
adds option to most implementations

naomijub:


naomijub:
to solve issue https://github.com/naomijub/edn-rs/issues/15

PR #21, opened 2020-07-25
Error type for Result

naomijub:


evaporei (approved):
Let's go to version 10 just like you :point_right: 
:100:

PR #23, opened 2020-08-03
Removes ordered float to hash f64 as (i64, u64).

naomijub:
Changes are only under the hood, objective is to remove dependency.
https://github.com/naomijub/edn-rs/issues/22


evaporei:
The solution seems fine, I would like to test some data just to make sure the conversion works properly.

evaporei (approved):
Noice :shipit:

PR #26, opened 2020-08-15
add futures to edn

naomijub:


evaporei:
First, are those println! suppose to be there?
Second, the `to_string` then `to_owned` doesn't make the string conversion happen twice for each case?
Third, if the string goes empty, will it ever get Ready? Isn't this an infinite loop?
Fourth, I don't see much value in this feature when the Edn parsing is synchronous, can you tell me more about the benefits? Is it all about making it more plugable with async code? Because sync code should be pretty plugable with async code as far as I know.

evaporei (approved):

PR #28, opened 2020-08-16
limits json/regex usage to feature json

naomijub:


tests/lib.rs
```diff
@@ -0,0 +1,7 @@
+#![recursion_limit = "512"]
```
evaporei, line 1: Is this necessary?

.github/workflows/rust.yml
```diff
@@ -11,11 +11,13 @@ jobs:
     - uses: actions/checkout@v1
     - name: Build
       run: cargo build --verbose
-    - name: Run tests
-      run: cargo test --verbose
+    - name: Run unit tests
+      run: cargo test --locked  --no-fail-fast --lib
+    - name: Run int tests
+      run: cargo test --test lib --no-fail-fast --features "json"
```
evaporei, line 17: Why is there a `lib` here? I am a bit confused since you're using `lib` here and `--lib` flag on `.travis.yml` :thinking:

tests/emit.rs
```diff
@@ -1,6 +1,6 @@
 extern crate edn_rs;
 
-#[cfg(test)]
+#[cfg(feature = "json")]
```
evaporei, line 3: Shouldn't it be both `json` feature and `test`?
We're probably compiling it even when its not test compilation.

Makefile
```diff
@@ -0,0 +1,7 @@
+int:
```
evaporei, line 1: What does `int` mean?

README.md
```diff
@@ -192,7 +194,7 @@ async fn main() {
 - [x] Define EDN types, `EdnType`
  - [x] Edn Type into primitive: `Edn::Bool(true).into() -> true`. This was done by `to_float`, `to_bool`, `to_int`, `to_vec`.
  - [x] implement `futures::Future` trait to `Edn`
-- [x] Parse EDN data [`parse_edn`](https://docs.rs/edn-rs/0.10.2/edn_rs/deserialize/fn.parse_edn.html):
+- [x] Parse EDN data [`from_str`](https://docs.rs/edn-rs/0.10.2/edn_rs/deserialize/fn.from_str.html):
```
evaporei, line 197: Is this changed because the branch needs a rebase or I forgot to do it on the other PR?

.github/workflows/rust.yml
naomijub: the lib file in the `test/` dir

tests/emit.rs
naomijub: test dir doesn't compile

Makefile
naomijub: integration, same as transistor, as test/ folder is a integration test to cargo

PR #31, opened 2020-08-16
Deserialize

evaporei:
A Pull Request just to sharpen a bit the `Deserialize` trait, and make `from_str` be able to resolve for any type which implements it.


.github/workflows/rust.yml
```diff
@@ -19,7 +19,8 @@ jobs:
       run: |
         cargo run --example json_to_edn --features "json"
         cargo run --example navigate_edn
-        cargo run --example parse_str
+        cargo run --example from_str_edn
+        cargo run --example from_str_type
```
naomijub: what about `struct_from_str` and `edn_from_str`?

evaporei: Fair enough!

README.md
```diff
@@ -67,13 +67,9 @@ fn main() -> Result<(), String> {
                         Edn::Map(Map::new(map!{":a".to_string() => Edn::Key(":b".to_string())})),
                         Edn::Key(":A".to_string()),
                         Edn::Nil}))}
-        )))
+        ))
     );
 
-    // OR 
-
-    let edn_resp = edn_rs::from_str(edn_str)?;
```
naomijub, line 75: the ideia of this test was to show navigation on the Edn response, I think it is important to have it in the readme

evaporei: Okay, I will do the same with the above structure.

README.md
```diff
@@ -128,6 +124,58 @@ fn main() {
...
+    age: usize,
+}
+
+impl Deserialize for Person {
+    fn deserialize(edn: Edn) -> Result<Self, EdnError> {
+        Ok(Self {
+            name: edn[":name"].to_string(),
+            age: edn[":age"].to_uint().ok_or_else(|| {
+                EdnError::Deserialize("couldn't convert `:age` into `uint`".to_string())
+            })?,
```
naomijub, line 147: I think there is a small problem when we developed, that the main type for integer is isize and no usize, just something to review.

evaporei: But I think isize should be default, since we don't know if the user is using negatives or not.

evaporei:
By the way, this is the default deserialization for primitive types that I was doing:

```rust
fn build_deserialize_error(edn: Edn, type_: &str) -> Error {
    Error::Deserialize(format!("couldn't convert `{}` to `{}`", edn, type_))
}

macro_rules! impl_deserialize_float {
    ( $( $name:ty ),+ ) => {
        $(
            impl Deserialize for $name
            {
                fn deserialize(edn: &Edn) -> Result<Self, Error> {
                    edn
                        .to_float()
                        .ok_or_else(|| build_deserialize_error(edn.clone(), "float"))
                        .map(|u| u as $name)
                }
            }
        )+
    };
}

impl_deserialize_float!(f32, f64);

macro_rules! impl_deserialize_int {
    ( $( $name:ty ),+ ) => {
        $(
            impl Deserialize for $name
            {
                fn deserialize(edn: &Edn) -> Result<Self, Error> {
                    edn
                        .to_int()
                        .ok_or_else(|| build_deserialize_error(edn.clone(), "int"))
                        .map(|u| u as $name)
                }
            }
        )+
    };
}

impl_deserialize_int!(isize, i8, i16, i32, i64);

macro_rules! impl_deserialize_uint {
    ( $( $name:ty ),+ ) => {
        $(
            impl Deserialize for $name
            {
                fn deserialize(edn: &Edn) -> Result<Self, Error> {
                    edn
                        .to_uint()
                        .ok_or_else(|| build_deserialize_error(edn.clone(), "uint"))
                        .map(|u| u as $name)
                }
            }
        )+
    };
}

impl_deserialize_uint!(usize, u8, u16, u32, u64);

impl Deserialize for bool {
    fn deserialize(edn: &Edn) -> Result<Self, Error> {
        edn.to_bool()
            .ok_or_else(|| build_deserialize_error(edn.clone(), "bool"))
    }
}

impl Deserialize for String {
    fn deserialize(edn: &Edn) -> Result<Self, Error> {
        Ok(edn.to_string())
    }
}
```

examples/from_str_edn.rs
```diff
@@ -3,25 +3,21 @@ use std::str::FromStr;
...
-                Edn::Nil}))})))
+                Edn::Nil}))}))
     );
 
     println!("{:?}", edn);
 
-    // OR
-
-    let edn_resp = edn_rs::from_str(edn_str)?;
-    assert_eq!(edn_resp[":b"][0], Edn::Bool(true));
```
naomijub: Navigation docs

src/deserialize/mod.rs
```diff
@@ -1,20 +1,64 @@
 use crate::edn::Error;
 use crate::edn::{Edn, List, Map, Set, Vector};
-use std::marker::Sized;
+use std::str::FromStr;
 /// public trait to be used to `Deserialize` structs
-/// A possible solution can be found [here](https://github.com/naomijub/transistor/blob/master/src/types/response.rs#L152)
+///
+/// Example:
```
naomijub, line 6: Nice!

src/deserialize/mod.rs
```diff
@@ -1,20 +1,64 @@
...
+/// assert_eq!(
+///     person,
+///     Err(EdnError::Deserialize(
+///         "couldn't convert `:age` into `uint`".to_string()
+///     ))
+/// );
+/// ```
 pub trait Deserialize: Sized {
-    fn deserialize<S>(self) -> Result<Self, Error>;
+    fn deserialize(edn: Edn) -> Result<Self, Error>;
```
naomijub, line 52: Nice!

.github/workflows/rust.yml
evaporei: Solved!

README.md
evaporei: Solved!

examples/from_str_edn.rs
evaporei: Solved!

src/deserialize/mod.rs
```diff
@@ -34,7 +78,7 @@ fn tokenize(edn: &str) -> Vec<String> {
         .collect()
 }
 
-fn parse<'a>(tokens: &'a [String]) -> Result<(Edn, &'a [String]), Error> {
+pub(crate) fn parse<'a>(tokens: &'a [String]) -> Result<(Edn, &'a [String]), Error> {
```
naomijub, line 81: Just a loose comment, I think that `parse_word` should return `Edn::Empty` instead of and Error if the string is empty, what do you think?

src/edn/mod.rs
```diff
@@ -516,7 +517,9 @@ impl std::str::FromStr for Edn {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        crate::deserialize::from_str(s)
+        let tokens = deserialize::tokenize(s);
```
naomijub, line 520: didn't understand?

src/edn/mod.rs
```diff
@@ -542,19 +545,21 @@ fn rational_to_double(r: &str) -> Option<f64> {
...
 #[derive(Debug, PartialEq)]
 pub enum Error {
-    ParseEdnError(String),
+    ParseEdn(String),
+    Deserialize(String),
 }
 
 impl From<String> for Error {
     fn from(s: String) -> Self {
-        Error::ParseEdnError(s)
```
naomijub, line 550: Most of our Errors finish with `..Error` in other crates, is this (remove `Error` from name) the standard of the community?

src/edn/mod.rs
evaporei: I move the code from old `from_str` to `Edn::from_str`, so now it tokenizes, than parse the string to an `Edn` here, just like before.

src/edn/mod.rs
naomijub: Also, what do you think of error handling like this? https://docs.serde.rs/src/serde/de/mod.rs.html#159

evaporei: I think it is redundant to have each variant with a `Error` suffix. I prefer without it, what do you think?

Example of redundancy:

```rust
EdnError::ParseEdnError(str)
EdnError::DeserializeError(str)
```

Instead of:

```rust
EdnError::ParseEdn(str)
EdnError::Deserialize(str)
```

Honestly I think I would even remove the `Edn` out of the `ParseEdn` variant

src/deserialize/mod.rs
naomijub: This is non blocker, maybe a issue

naomijub (approved):

src/deserialize/mod.rs
evaporei: Hmmm, maybe that's a good idea actually, I will think about a little than I respond you :slightly_smiling_face:

naomijub:
Bump version! `0.11.4`

evaporei:
@naomijub honestly I was thinking of bumping to `0.12.0` since it breaks the `Deserialize` interface, and documents how you should do deserialization using the trait. What do you think?

src/edn/mod.rs
evaporei: Wow, that seems complex haha, I need sometime to undertand that :joy:

src/deserialize/mod.rs
evaporei: Hmm okay, why do we have `Edn::Empty`? In the spec does it say something about an empty string being valid `edn`?

If there's nothing on it about it, I think we should remove this variant from `Edn` enum.

src/edn/mod.rs
evaporei: I will resolve this, maybe I will look at this later, thanks for pointing out something that we can be based of in the future :)

PR #34, opened 2020-08-22
implements various to_type_vec

naomijub:


evaporei:
I don't like this conversion because it omits values that are on the list which aren't the same type.
IMO that's obscure and it is unfortunately the kind of conversion the user will have to do.

What I prefer is, if a list has all members of the same type, then the conversion works, instead of returning a filtered list.

What do you think?

Edit: I just saw your other [PR](https://github.com/naomijub/edn-rs/pull/35) which does exactly what I've said, so I will me merging that one, than we can keep discussing later about this one :wink:

PR #36, opened 2020-08-22
Deserialize primitives

evaporei:
This will not only make the user have less trouble implementing `Deserialize`, but also will help us out on `edn-derive` for generic implementations over any type :tada:

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,12 +139,10 @@ struct Person {
 }
 
 impl Deserialize for Person {
-    fn deserialize(edn: Edn) -> Result<Self, EdnError> {
+    fn deserialize(edn: &Edn) -> Result<Self, EdnError> {
         Ok(Self {
-            name: edn[":name"].to_string(),
-            age: edn[":age"].to_uint().ok_or_else(|| {
-                EdnError::Deserialize("couldn't convert `:age` into `uint`".to_string())
-            })?,
+            name: Deserialize::deserialize(&edn[":name"])?,
+            age: Deserialize::deserialize(&edn[":age"])?,
         })
     }
 }
@@ -164,13 +162,13 @@ fn main() -> Result<(), EdnError> {
     println!("{:?}", person);
     // Person { name: "rose", age: 66 }
 
-    let bad_edn_str = "{:name \"rose\" :age \"not an uint\"}";
+    let bad_edn_str = "{:name \"rose\" :age \"some text\"}";
     let person: Result<Person, EdnError> = edn_rs::from_str(bad_edn_str);
 
     assert_eq!(
         person,
         Err(EdnError::Deserialize(
-            "couldn't convert `:age` into `uint`".to_string()
+            "couldn't convert `some text` to `uint`".to_string()
         ))
     );
 
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ fn main() -> Result<(), EdnError> {
     assert_eq!(
         person,
         Err(EdnError::Deserialize(
-            "couldn't convert `some text` to `uint`".to_string()
+            "couldn't convert `some text` into `uint`".to_string()
         ))
     );
 
--- a/examples/complex_struct_deserialization.rs
+++ b/examples/complex_struct_deserialization.rs
@@ -66,7 +66,7 @@ fn main() -> Result<(), EdnError> {
     assert_eq!(
         complex,
         Err(EdnError::Deserialize(
-            "couldn't convert `some text` to `uint`".to_string()
+            "couldn't convert `some text` into `uint`".to_string()
         ))
     );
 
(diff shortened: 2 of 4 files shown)


evaporei:
Here's an example of how `edn-derive` becomes simpler and more generic :tada: 
https://github.com/otaviopace/edn-derive/pull/5

README.md
```diff
@@ -164,13 +162,13 @@ fn main() -> Result<(), EdnError> {
...
 
-    let bad_edn_str = "{:name \"rose\" :age \"not an uint\"}";
+    let bad_edn_str = "{:name \"rose\" :age \"some text\"}";
     let person: Result<Person, EdnError> = edn_rs::from_str(bad_edn_str);
 
     assert_eq!(
         person,
         Err(EdnError::Deserialize(
-            "couldn't convert `:age` into `uint`".to_string()
+            "couldn't convert `some text` to `uint`".to_string()
```
naomijub: `:age \"some text\" into`?

examples/complex_struct_deserialization.rs
```diff
@@ -0,0 +1,74 @@
+use edn_rs::{Deserialize, Edn, EdnError};
+
+#[derive(Debug, PartialEq)]
```
naomijub, line 3: Do you think it makes sense to have an example of `Deserialize` macro here?

src/deserialize/mod.rs
```diff
@@ -38,24 +36,128 @@ use std::str::FromStr;
...
 ///     ))
 /// );
 /// ```
 pub trait Deserialize: Sized {
-    fn deserialize(edn: Edn) -> Result<Self, Error>;
+    fn deserialize(edn: &Edn) -> Result<Self, Error>;
+}
+
+fn build_deserialize_error(edn: Edn, type_: &str) -> Error {
+    Error::Deserialize(format!("couldn't convert `{}` to `{}`", edn, type_))
```
naomijub: **convert into/convert to**
Convert into means to change from one thing to another; 
Convert to means to switch allegiance, loyalty, or obligation.

So `into`

src/deserialize/mod.rs
```diff
@@ -38,24 +36,128 @@ use std::str::FromStr;
...
+                        .to_uint()
+                        .ok_or_else(|| build_deserialize_error(edn.clone(), "uint"))
+                        .map(|u| u as $name)
+                }
+            }
+        )+
+    };
+}
+
+impl_deserialize_uint!(usize, u8, u16, u32, u64);
```
naomijub, line 109: MAYBE  we can implement without this macros? something like:

```
match edn {
  e if e.to_int().is_some() => e.to_uint().ok_or_else(|| build_deserialize_error(edn.clone(), "int")).map(|u| u as isize),
  e if e.to_float().is_some() => e.to_float().ok_or_else(|| build_deserialize_error(edn.clone(), "float")).map(|u| u as f64),
...
}
```

naomijub: Also, maybe we will have a problem for types that are not the ones that we use with Edn?

src/deserialize/mod.rs
```diff
@@ -38,24 +36,128 @@ use std::str::FromStr;
...
+        edn.to_char()
+            .ok_or_else(|| build_deserialize_error(edn.clone(), "char"))
+    }
+}
+
+impl<T> Deserialize for Vec<T>
+where
+    T: Deserialize,
+{
+    fn deserialize(edn: &Edn) -> Result<Self, Error> {
```
naomijub, line 135: Nice!

README.md
evaporei: I've changed the `bad_edn_str` to contain `"some text"`, I think it seems more explicit. It's just a string value change, no behavioral change.

examples/complex_struct_deserialization.rs
evaporei: I thought about that. I think we should add once this is on master, and the `edn-derive` is using it :slightly_smiling_face: 
I can do that PR, no probs.

src/deserialize/mod.rs
evaporei: Okay!

src/deserialize/mod.rs
evaporei: Yeah! I think we can follow for now just with `Vec<T>`, but we definitely will need other types like `BTreeMap`, `BTreeSet`, and so on and so forth.

src/deserialize/mod.rs
evaporei: I prefer macros here because it will be faster at runtime, even though the difference is minimal.

evaporei: For external types, the user should implement `Deserialize` for it, then the other types that use it can use `derive` for example :)

naomijub (approved):

PR #37, opened 2020-08-24
Add default Deserialization for Option<T>

evaporei:
This deserialization will help on types that use `Option`, for example if someone makes a struct with a field `Option<CruxId>`, this will make it work :slightly_smiling_face:


naomijub (approved):
Does it have tests?

evaporei:
Not yet, I will add some then I will ask for another review :slightly_smiling_face:

evaporei:
Maybe its done now! :thinking:

PR #40, opened 2020-08-30
Improves performance EDN parse by ~80% by using std::str::Chars

naomijub:
```
Benchmarking parse: Collecting 100 samples in estimated 5.0323 s (439k iteration                                                                              
parse                   time:   [13.570 us 13.606 us 13.641 us]
Performance has improved: 80.560%
```

We still can improve to 5us

PLEASE REVIEW BUT DONT MERGE IT


Makefile
```diff
@@ -4,4 +4,9 @@ int:
...
 	cargo test --locked  --no-fail-fast --lib
 
-test: unit int
\ No newline at end of file
+ex:
+	cargo test --examples
+	cargo test --example json_to_edn --features "json"
+	cargo run --example async --features "async"
+
+test: unit int ex
```
evaporei: What do you think of we using `integration` and `examples` instead of `int` and `ex` on the `Makefile`?

We can use `.PHONY: examples` for example, to don't clash with file names :slightly_smiling_face:

naomijub: Third time is the charm 33a4851

examples/complex_struct_deserialization.rs
```diff
@@ -72,3 +75,19 @@ fn main() -> Result<(), EdnError> {
...
+
+fn main() -> Result<(), EdnError> {
+    complex_ok()?;
+    complex_wrong()?;
+    Ok(())
+}
+
+#[test]
+fn test_complex_ok() {
+    let _ = complex_ok();
```
evaporei, line 87: I don't get these new tests on the examples, isn't this duplicate code?

Also, if it isn't, shouldn't this be unwrapping or asserting it is an Ok value? Because the way it is, it seem to be ignoring the result, and the test could never fail

naomijub: The `main` function is for the `cargo run --examples <example_name>` while `test` is for the `cargo test --examples`. We run `cargo test --examples` in the CI while a user can run `cargo run --examples <example_name>` to understand the example

naomijub: The Result of the function call doesn't ignore the assert, it just happens before the return, so if the test fails the example will fail as well.

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+            res.insert(key.unwrap().to_string(), val.unwrap());
+            key = None;
+            val = None;
+        }
+    }
+}
+
+use std::borrow::Cow;
+
+pub trait MaybeReplaceExt<'a> {
```
naomijub, line 189: improves performance by `3 us`

src/deserialize/mod.rs
```diff
@@ -136,17 +140,17 @@ where
         match edn {
             Edn::Vector(_) => Ok(edn
                 .iter()
-                .unwrap()
+                .ok_or(Error::Iter(format!("Could not create iter from {:?}", edn)))?
```
evaporei, line 141: Wow what a great addition!!! :clap:

src/deserialize/parse.rs
evaporei: Is this trait necessary though? If you convert an `&str` to a `Cow` and do the default Rust `.replace` method, wouldn't it only change if the replace actually works?

I mean, the `Cow` stands for Copy On Write as far as I know, doesn't is support native checking for this? :thinking:

src/deserialize/mod.rs
```diff
@@ -171,171 +175,19 @@ where
 
 /// `from_str` deserializes an EDN String into type `T` that implements `Deserialize`. Response is `Result<T, EdnError>`
 pub fn from_str<T: Deserialize>(s: &str) -> Result<T, Error> {
-    let edn = Edn::from_str(s)?;
+    let clean = String::from(s.maybe_replace("#{", "@"));
```
evaporei: I think this replace is duplicate, you already do it on `Edn::from_str`.

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+}
+
+pub(crate) fn parse_edn(c: Option<char>, chars: &mut std::str::Chars) -> Result<Edn, Error> {
+    match c {
+        Some('\"') => Ok(read_str(chars)),
+        Some(':') => Ok(read_key(chars)),
+        Some(n) if n.is_numeric() => Ok(read_number(n, chars)?),
+        Some('-') => Ok(read_number('-', chars)?),
+        Some('\\') => Ok(read_char(chars)?),
+        Some(b) if b == 't' || b == 'f' || b == 'n' => Ok(read_bool_or_nil(b, chars)?),
```
evaporei, line 23: Can't this be `Some('t') | Some('f') | Some('n')`?

I think ifs like this on match clauses are probably slower than the regular's Rust pattern matching (idk the internals, I might be wrong, but I think it is worth the try).

Also I think it is cleaner.

If you're worried about using the `c` variable later, you can use the match binding (`@`) like this:

```rust
Some(b) @ Some('t') | Some('f') | Some('n') => Ok(read_bool_or_nil(b, chars)?)
```

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+        a => Err(Error::ParseEdn(format!(
+            "{} could not be parsed",
+            a.unwrap().to_string()
```
evaporei: Is this sure to be an `Ok` value?

Maybe it is more verbose, but we can divide this match arm into two, one where `a` is `Ok(...)` and other where it is `Err(...)`.

Then the `unwrap` is unnecessary and we can use the error that came from `Err`

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+        Some('{') => read_map(chars)?,
+        edn => parse_edn(edn, chars)?,
+    })
+}
+
+pub(crate) fn parse_edn(c: Option<char>, chars: &mut std::str::Chars) -> Result<Edn, Error> {
+    match c {
+        Some('\"') => Ok(read_str(chars)),
+        Some(':') => Ok(read_key(chars)),
+        Some(n) if n.is_numeric() => Ok(read_number(n, chars)?),
```
evaporei: Have you tried putting this match arm one before the error one? I tend to think ifs inside match clauses are slower than pure pattern matches, I believe Rust optimize those somehow (again, might be wrong).

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+        .collect::<String>()
+        .len();
+    let mut key = String::from(":");
+    let key_chars = chars.take(c_len).collect::<String>();
+    key.push_str(&key_chars);
+    Edn::Key(key)
+}
+
+fn read_str(chars: &mut std::str::Chars) -> Edn {
+    let string = chars.take_while(|c| c != &'\"').collect::<String>();
```
evaporei, line 44: What happens if there's two backslashes (`\`) inside a String? :thinking:

evaporei: Actually thinking now you're looking for the `\"` character, so it probably isn't a problem what I am saying.

src/deserialize/parse.rs
evaporei: Reference: https://doc.rust-lang.org/stable/rust-by-example/flow_control/match/binding.html

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+
+fn read_vec(chars: &mut std::str::Chars) -> Result<Edn, Error> {
+    let mut res: Vec<Edn> = vec![];
+    loop {
+        match chars.next() {
+            Some(']') => return Ok(Edn::Vector(Vector::new(res))),
+            Some(c) if !c.is_whitespace() && c != ',' => {
+                res.push(parse(Some(c), chars)?);
+            }
+            Some(c) if c.is_whitespace() || c == ',' => (),
```
evaporei, line 125: I think we can make those two `Some(c)` arms faster by calling `is_whitespace` and checking if the character is a `,` only once.

Maybe something like this:
```rust
Some(c) => {
    let is_comma = c == ',';
    if !c.is_whitespace() && !is_comma {
        res.push(parse(Some(c), chars)?);
    }
}
```

Another option is to:
```rust
Some(c) if !c.is_whitespace() && c != ',' => {
    res.push(parse(Some(c), chars)?);
}
```
And to just remove the second `Some(c)` since it does nothing :wink:

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+                .take_while(|e| e == &'i' || e == &'l')
+                .collect::<String>()
+                .len();
+            let mut string = String::new();
+            let n = chars.take(c_len).collect::<String>();
+            string.push(c);
+            string.push_str(&n);
+            match &string[..] {
+                "nil" => Ok(Edn::Nil),
+                _ => Err(Error::ParseEdn(format!("{} cound not be parsed", string))),
```
evaporei: Typo, it should be `could`.

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+            }
+        }
+        _ => Err(Error::ParseEdn(
+            "Nullable boolean cound not be parsed".to_string(),
```
evaporei: Typo, it should be `could`

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+
+fn read_list(chars: &mut std::str::Chars) -> Result<Edn, Error> {
+    let mut res: Vec<Edn> = vec![];
+    loop {
+        match chars.next() {
+            Some(')') => return Ok(Edn::List(List::new(res))),
+            Some(c) if !c.is_whitespace() && c != ',' => {
+                res.push(parse(Some(c), chars)?);
+            }
+            Some(c) if c.is_whitespace() || c == ',' => (),
```
evaporei, line 139: Again, I think this second `Some(c)` is not doing anything, it could be removed.

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+fn read_set(chars: &mut std::str::Chars) -> Result<Edn, Error> {
+    use std::collections::BTreeSet;
+    let mut res: BTreeSet<Edn> = BTreeSet::new();
+    loop {
+        match chars.next() {
+            Some('}') => return Ok(Edn::Set(Set::new(res))),
+            Some(c) if !c.is_whitespace() && c != ',' => {
+                res.insert(parse(Some(c), chars)?);
+            }
+            Some(c) if c.is_whitespace() || c == ',' => (),
```
evaporei, line 154: Can't this be removed like the other `Some(c)`? :thinking:

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+        match chars.next() {
+            Some('}') => return Ok(Edn::Map(Map::new(res))),
+            Some(c) if !c.is_whitespace() && c != ',' => {
+                if key.is_some() {
+                    val = Some(parse(Some(c), chars)?);
+                } else {
+                    key = Some(parse(Some(c), chars)?);
+                }
+            }
+            Some(c) if c.is_whitespace() || c == ',' => (),
```
evaporei, line 175: Maybe removal of unnecessary arm? :thinking:

src/deserialize/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+
+    #[test]
+    fn parse_char() {
+        let mut c = "\\k".chars();
+
+        assert_eq!(parse_edn(c.next(), &mut c).unwrap(), Edn::Char('k'))
+    }
+
+    #[test]
+    fn parse_bool_or_nil() {
```
evaporei, line 265: I think we should test strings like `ture` and `flase` to check if it returns an error properly.

evaporei:
Great PR :clap: :clap: :clap: :clap: :clap:
Congrats on making the parsing more reliable!! :blush: 
I just added a couple notes on some things I think can be influencing performance, can be simplified and or could cause problems :wink:

src/deserialize/parse.rs
naomijub: it improved performance in 20%

src/deserialize/parse.rs
naomijub: test result: FAILED. 36 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out

src/deserialize/parse.rs
naomijub: failures:
    deserialize::parse::test::parse_complex
    deserialize::parse::test::parse_set
    deserialize::test::from_str_complex_map
    deserialize::test::from_str_list_with_set

src/deserialize/parse.rs
naomijub: failures:
    deserialize::parse::test::parse_complex
    deserialize::parse::test::parse_list
    deserialize::test::from_str_list_with_set
    deserialize::test::from_str_list_with_vec

src/deserialize/parse.rs
naomijub: You cant remove the second one because he is responsible for skipping values.

naomijub: they both should exist because it seems important to have an `err` case. If we get something unknown we should return an error and not just skip it.

src/deserialize/parse.rs
naomijub: It will look like this: 
```b @ Some('t') | b @ Some('f') | b @ Some('n') => Ok(read_bool_or_nil(b.unwrap(), chars)?),```

Comment (naomijub) [reply src/deserialize/parse.rs:23]: With binding:
`[14.173 us 14.470 us 14.756 us]`

With if:
`[13.232 us 13.264 us 13.303 us]`

File: src/deserialize/parse.rs (hunk 10 above)
Comment (naomijub) [reply src/deserialize/parse.rs:125]: Tried replacing for
```rust
const fn should_skip_char(c: &char) -> bool {
    *c == ' ' || *c == ',' || *c == '\n' || *c == '\r' || *c == '\t'
}
```

and this:

```rust
const fn should_skip_char(c: &char) -> bool {
     match *c {
         ' ' | ',' | '\x09'..='\x0d' => true,
         _ => false
     }
}
```

And it only worsen performance

naomijub: This also worsened performance:
```rust
let ch = chars.next();
        let should_skip = ch.unwrap_or(' ').is_whitespace() || ch.unwrap_or(' ') == ',';
        match ch {
            Some('}') => return Ok(Edn::Map(Map::new(res))),
            Some(c) if !should_skip => {
                if key.is_some() {
                    val = Some(parse(Some(c), chars)?);
                } else {
                    key = Some(parse(Some(c), chars)?);
                }
            }
            Some(_) if should_skip => (),
            err => return Err(Error::ParseEdn(format!("{:?} could not be parsed", err))),
        }
```

naomijub: This actually worsen performance by 2us:
`[14.930 us 15.222 us 15.555 us]`
```rust
match c.next() {
            Some(']') => return Ok(Edn::Vector(Vector::new(res))),
            Some(c) => {
                let should_skip = c.is_whitespace() || c == ',';
                if !should_skip {
                    res.push(parse(Some(c), chars)?);
                }
            }
}```

Comment (evaporei) [reply src/deserialize/parse.rs:125]: How about using this if then on the `Err` arm, like this:

```rust
err if !(c.is_whitespace() || c == ',') => return Err(Error::ParseEdn(format!("{:?} could not be parsed", err))),
```

src/deserialize/parse.rs
evaporei: Ohh that's sad :slightly_frowning_face:

src/deserialize/parse.rs
evaporei: Okay, I thought more, isn't this a problem when the character before this one is a `\`, because it's trying to skip the next slash? :thinking:

naomijub: if the char before is a `\` its char representation will be `\\` not `\`

src/deserialize/parse.rs
naomijub: it doesn't improve performance as the other cases show

src/deserialize/parse.rs
naomijub: abd451e

src/deserialize/parse.rs
evaporei: I actually meant that the `err` should have this logic, but it's the same as this.

evaporei (approved):

PR #42, opened 2020-09-09
#Inst as Edn::Inst

naomijub:
This PR:
- [x] type Edn::Inst

Next PR:
- [ ] to_datetime_utc, with chrono
- [ ] to_datetime_fixedtime, with chrono

### DIFF
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -47,7 +47,9 @@ fn read_str(chars: &mut std::str::Chars) -> Edn {
 }
 
 fn read_inst(chars: &mut std::str::Chars) -> Result<Edn, Error> {
-    let inst = chars.take_while(|c| c != &'\"').collect::<String>();
+    let inst = chars
+        .take_while(|c| c != &'\"' || (*c).is_numeric())
+        .collect::<String>();
     let time = chars.take_while(|c| c != &'\"').collect::<String>();
 
     if inst.contains("inst") {
@@ -400,5 +402,14 @@ mod test {
                 Edn::Nil
             }))
         )
+    }
+
+    #[test]
+    #[should_panic(expected = "iasdf 234  could not be parsed")]
+    fn panic_for_wrong_inst() {
+        let mut edn = "#iasdf 234".chars();
+        let a = parse(edn.next(), &mut edn);
+
+        a.unwrap();
     }
 }
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -52,7 +52,7 @@ fn read_inst(chars: &mut std::str::Chars) -> Result<Edn, Error> {
         .collect::<String>();
     let time = chars.take_while(|c| c != &'\"').collect::<String>();
 
-    if inst.contains("inst") {
+    if inst.starts_with("inst") {
         return Ok(Edn::Inst(time));
     }
     Err(Error::ParseEdn(format!(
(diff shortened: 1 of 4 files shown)


src/deserialize/parse.rs
```diff
@@ -45,6 +46,21 @@ fn read_str(chars: &mut std::str::Chars) -> Edn {
     Edn::Str(string)
 }
 
+fn read_inst(chars: &mut std::str::Chars) -> Result<Edn, Error> {
+    let inst = chars
+        .take_while(|c| c != &'\"' || (*c).is_numeric())
+        .collect::<String>();
+    let time = chars.take_while(|c| c != &'\"').collect::<String>();
+
+    if inst.contains("inst") {
```
evaporei: Shoudn't this be more precise instead of `contains`? :thinking:

naomijub: just improved a little. What do you suggest? `!= "inst"`?

naomijub: If we use ` .take_while(|c| "inst ".contains(*c))` the error doesn't make sense:
```
 panic message: `"called `Result::unwrap()` on an `Err` value: ParseEdn(\"isdf 234 could not be parsed\")"`,
 expected substring: `"iasdf 234  could not be parsed"`
```
For test:
```rust
#[test]
#[should_panic(expected = "iasdf 234  could not be parsed")]
fn panic_for_wrong_inst() {
    let mut edn = "#iasdf 234".chars();
    let a = parse(edn.next(), &mut edn);

    a.unwrap();
}
```

evaporei: What I mean by precise is that I guess someone can pass something like `#instantenous "..."` or `#blablainst "..."`

Can't we do `starts_with` and then check that the rest is just extra spaces? :thinking:

naomijub: Ah OK! makes sense

naomijub: Will fix

naomijub: 0f8f94d

evaporei: So there's still the problem with `instanteneous` for example. I've cloned the branch and created this test, but it should not pass, and it passes:

```rust
    #[test]
    fn parse_edn_with_instanteneous() {
        let mut edn =
            "@ :a :b {:c :d :date  #instanteneous \"2020-07-16T21:53:14.628-00:00\" ::c ::d} nil}".chars();

        assert_eq!(
            Edn::Set(Set::new(set! {
                Edn::Key(":a".to_string()),
                Edn::Key(":b".to_string()),
                Edn::Map(Map::new(map! {
                    ":c".to_string() => Edn::Key(":d".to_string()),
                    ":date".to_string() => Edn::Inst("2020-07-16T21:53:14.628-00:00".to_string()),
                    "::c".to_string() => Edn::Key("::d".to_string())
                })),
                Edn::Nil
            }))
        )
    }
```

Sorry to not bring a solution, I am a bit busy.

naomijub: I know we talked, but as we don't have a solution now I am answering here what we talked:
`This lib does not make effort to conform the EDN received to EDN Spec. The lib that generated this EDN should be responsible for this. For more information on Edn Spec please visit: https://github.com/edn-format/edn.`

When we find a solution we can change this problem

evaporei: Okay then :confused:

evaporei (approved):

PR #45, opened 2020-09-11
Start Trailing spaces

naomijub:
`cargo run --example struct_from_str`
 returns
`Error: ParseEdn("  could not be parsed")`


evaporei:
What is the reason for this change?

evaporei (approved):

PR #47, opened 2020-09-12
Fix keyword

naomijub:


evaporei:
I've created this test with the same implementation, but it fails:
```rust
#[test]
fn parse_map_keyword_with_commas() {
    let mut edn = "{ :a :something, :b false, :c nil, }".chars();

    assert_eq!(
        Edn::Map(Map::new(
            map! {":a".to_string() => Edn::Key("something".to_string()),
            ":b".to_string() => Edn::Bool(false), ":c".to_string() => Edn::Nil}
        ))
    );
}
```

naomijub (commented):
From my point of view the test should be:
```rust
#[test]
    fn parse_map_keyword_with_commas() {
        let mut edn = "{ :a :something, :b false, :c nil, }".chars();

        assert_eq!(
            Edn::Map(Map::new(
                map! {":a".to_string() => Edn::Key(":something".to_string()),
                ":b".to_string() => Edn::Bool(false), ":c".to_string() => Edn::Nil}
            ))
        );
    }
```

src/edn/mod.rs
```diff
@@ -223,7 +223,7 @@ impl core::fmt::Display for Vector {
             "[{}]",
             self.0
                 .iter()
-                .map(|i| format!("{:?}, ", i))
+                .map(|i| format!("{}, ", i))
```
evaporei, line 226: I don't think I understood why those were removed?

Also, this file is using lots of `Edn::Key` without a `:` prefix on the `String`, we should add that to make the use consistent

naomijub: because this was including Edn::Type in the Display. So, for type `Edn::Key(:something)`: 

* `.map(|i| format!("{:?}, ", i))` becomes `Edn::Key(:something)`
* `.map(|i| format!("{}, ", i))` becomes `:something`

naomijub: if we remove the `:` `.map(|i| format!("{}, ", i))` becomes `something`, which would be wrong

PR #53, opened 2020-09-17
Namespaced Map

naomijub:
`:abc{:a :b :c :d}`

### DIFF
--- a/examples/async.rs
+++ b/examples/async.rs
@@ -9,7 +9,7 @@ async fn foo() -> impl Future<Output = Edn> + Send {
 async fn main() {
     let edn = foo().await.await;
 
-    println!("{}", edn.to_string());
+    assert_eq!("[1, 1.5, \"hello\", :key, ]", edn.to_string());
     assert_eq!(edn, edn!([1 1.5 "hello" :key]));
 
     assert_eq!(edn[1].to_float(), Some(1.5f64));
--- a/src/deserialize/mod.rs
+++ b/src/deserialize/mod.rs
@@ -285,4 +285,20 @@ mod test {
             )]))
         )
     }
+
+    #[test]
+    fn namespaced_maps() {
+        let edn = ":abc{ 0 :val 1 :value}";
+
+        assert_eq!(
+            Edn::from_str(edn).unwrap(),
+            Edn::NamespacedMap(
+                "abc".to_string(),
+                Map::new(map! {
+                    "0".to_string() => Edn::Key(":val".to_string()),
+                    "1".to_string() => Edn::Key(":value".to_string())
+                })
+            )
+        );
+    }
 }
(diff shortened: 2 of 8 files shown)


examples/async.rs
```diff
@@ -9,7 +9,7 @@ async fn foo() -> impl Future<Output = Edn> + Send {
 async fn main() {
     let edn = foo().await.await;
 
-    println!("{}", edn.to_string());
+    assert_eq!("[1, 1.5, \"hello\", :key, ]", edn.to_string());
```
evaporei, line 12: Should this be here? :thinking:

src/deserialize/mod.rs
```diff
@@ -289,4 +289,20 @@ mod test {
             )]))
         )
     }
+
+    #[test]
+    fn namespaced_maps() {
+        let edn = ":abc{ 0 :val 1 :value}";
```
evaporei, line 295: Have you tested with a space between `c` and `{`?

examples/async.rs
naomijub: it was a failing test

src/deserialize/parse.rs
```diff
@@ -30,11 +30,19 @@ pub(crate) fn parse_edn(c: Option<char>, chars: &mut std::str::Chars) -> Result<
     }
 }
 
-fn read_key(chars: &mut std::str::Chars) -> Edn {
-    let c_len = chars
+fn read_key_or_nsmap(chars: &mut std::str::Chars) -> Result<Edn, Error> {
+    let mut key_chars = chars
         .clone()
```
evaporei, line 35: Why is this clone here again? I mean, there's a new one down there :thinking:

src/deserialize/mod.rs
naomijub: It cannot have the space, checked

src/deserialize/parse.rs
naomijub: because `take_while` takes the value it matched wrongly

naomijub: All take_whiles are inside clones

src/deserialize/mod.rs
evaporei: But what does the spec say? I know we don't verify and return errors, but we should accept every valid EDN.

src/deserialize/parse.rs
evaporei: Yeah, but there's a new one on the function now, can we remove one of those, or is it necessary?

naomijub: it is necessary, and we removed from here 
```rust
fn read_key(chars: &mut std::str::Chars, c_len: usize) -> Edn {
    let mut key = String::from(":");
    let key_chars = chars.take(c_len).collect::<String>();
    key.push_str(&key_chars);
    Edn::Key(key)
}
```

src/deserialize/mod.rs
naomijub: I checked the spec

naomijub: What could happen is this `#myapp/Person {:first "Fred" :last "Mertz"}`. It is a different case, tagged map, and it is not really a valid edn, is similar to `#inst` and `#uuid` but needs a reader implemented in the client

evaporei: oh shoot. Okay then

yeah, it is way more featured than JSON.

evaporei (approved):

PR #56, opened 2020-09-18
Support Rust unit (de)serialization

evaporei:


src/deserialize/mod.rs
```diff
@@ -56,6 +56,15 @@ fn build_deserialize_error(edn: &Edn, type_: &str) -> Error {
     Error::Deserialize(format!("couldn't convert `{}` into `{}`", edn, type_))
 }
 
+impl Deserialize for () {
+    fn deserialize(edn: &Edn) -> Result<Self, Error> {
+        match edn {
+            Edn::Nil => Ok(()),
+            _ => Err(build_deserialize_error(&edn, "()")),
```
evaporei: We can use `unit` instead of `()` in the error message, I don't know which is clearer.

naomijub: `unit` seems more explicit than `()`

evaporei: Fixed!

PR #60, opened 2020-09-26
Reinclude symbol parsing

naomijub:


naomijub:
from issue https://github.com/naomijub/edn-rs/issues/59

src/deserialize/parse.rs
```diff
@@ -54,6 +52,23 @@ fn read_str(chars: &mut std::str::Chars) -> Edn {
     Edn::Str(string)
 }
 
+fn read_others(a: char, chars: &mut std::str::Chars) -> Result<Edn, Error> {
```
evaporei: I don't know if I like the `others` suffix :thinking:

src/deserialize/parse.rs
```diff
@@ -54,6 +52,23 @@ fn read_str(chars: &mut std::str::Chars) -> Edn {
...
+    let c_len = chars
+        .clone()
+        .enumerate()
+        .take_while(|&(i, c)| i <= 200 && !c.is_whitespace() && c != ')' && c != '}' && c != ']')
```
evaporei, line 59: Why the 200 here?

src/deserialize/parse.rs
naomijub: c49169c

src/deserialize/parse.rs
naomijub: it is like a `timeout`

src/deserialize/parse.rs
```diff
@@ -54,6 +52,23 @@ fn read_str(chars: &mut std::str::Chars) -> Edn {
...
+        .count();
+    match a {
+        ws if ws.is_whitespace() => Err(Error::ParseEdn(format!("\"{}\" could not be parsed", ws))),
+        c => {
+            let mut symbol = String::from(c);
+            let symbol_chars = chars.take(c_len).collect::<String>();
+            symbol.push_str(&symbol_chars);
+            Ok(Edn::Symbol(symbol))
+        }
+    }
```
evaporei: (style suggestion)
Do you think we will have more cases?
What do you think of 

```rust
if a.is_whitespace() {
    return Err(Error::ParseEdn(format!("\"{}\" could not be parsed", ws)));
}
let mut symbol = String::from(c);
let symbol_chars = chars.take(c_len).collect::<String>();
symbol.push_str(&symbol_chars);
Ok(Edn::Symbol(symbol))
```

evaporei: Actually `c` and `c_len` variables don't exist on my example.

naomijub: this was my first idea, but then we changed to `read_symbol` and it makes no more sense to have a match...

naomijub: 92520e8

src/deserialize/parse.rs
naomijub: 92520e8

evaporei (approved):
:shipit:

PR #67, opened 2020-10-11
Deserialize for std::collections

naomijub:
adds `#[deprecated]` to `ser_struct!`

### DIFF
--- a/src/deserialize/mod.rs
+++ b/src/deserialize/mod.rs
@@ -84,6 +84,14 @@ macro_rules! impl_deserialize_float {
 
 impl_deserialize_float!(f32, f64);
 
+impl Deserialize for crate::Double {
+    fn deserialize(edn: &Edn) -> Result<Self, Error> {
+        edn.to_float()
+            .ok_or_else(|| build_deserialize_error(&edn, "edn_rs::Double"))
+            .map(|u| u.into())
+    }
+}
+
 macro_rules! impl_deserialize_int {
     ( $( $name:ty ),+ ) => {
         $(
@@ -197,7 +205,11 @@ where
                 .ok_or_else(|| Error::Iter(format!("Could not create iter from {:?}", edn)))?
                 .map(|(key, e)| {
                     (
-                        ns.to_string() + "/" + key,
+                        if ns.starts_with(":") {
+                            ns.to_string() + "/" + key
+                        } else {
+                            String::from(":") + ns + "/" + key
+                        },
                         Deserialize::deserialize(e).unwrap(),
                     )
                 })
@@ -232,7 +244,11 @@ where
                 .ok_or_else(|| Error::Iter(format!("Could not create iter from {:?}", edn)))?
                 .map(|(key, e)| {
                     (
-                        ns.to_string() + "/" + key,
+                        if ns.starts_with(":") {
+                            ns.to_string() + "/" + key
+                        } else {
+                            String::from(":") + ns + "/" + key
+                        },
                         Deserialize::deserialize(e).unwrap(),
                     )
                 })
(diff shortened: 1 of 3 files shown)


src/deserialize/mod.rs
```diff
@@ -177,6 +186,128 @@ where
...
+impl<T: std::cmp::Eq + std::hash::Hash> Deserialize for HashSet<T>
+where
+    T: Deserialize,
+{
+    fn deserialize(edn: &Edn) -> Result<Self, Error> {
+        match edn {
+            Edn::Set(_) => Ok(edn
+                .set_iter()
+                .ok_or_else(|| Error::Iter(format!("Could not create iter from {:?}", edn)))?
+                .map(|e| Deserialize::deserialize(e).unwrap())
```
evaporei, line 276: I think we should favor `from_edn` instead of `Deserialize::deserialize`

evaporei: Also, shouldn't we return the possible errors too? This seems unsafe

src/deserialize/mod.rs
```diff
@@ -177,6 +186,128 @@ where
...
+impl<T> Deserialize for BTreeMap<String, T>
+where
+    T: Deserialize,
+{
+    fn deserialize(edn: &Edn) -> Result<Self, Error> {
+        match edn {
+            Edn::Map(_) => Ok(edn
+                .map_iter()
+                .ok_or_else(|| Error::Iter(format!("Could not create iter from {:?}", edn)))?
+                .map(|(key, e)| (key, Deserialize::deserialize(e).unwrap()))
```
evaporei, line 237: Again, `from_edn` > `Deserialize::deserialize` and also unsafety of using `unwrap`.

src/deserialize/mod.rs
```diff
@@ -177,6 +186,128 @@ where
...
+                .map_iter()
+                .ok_or_else(|| Error::Iter(format!("Could not create iter from {:?}", edn)))?
+                .map(|(key, e)| {
+                    (
+                        if ns.starts_with(":") {
+                            ns.to_string() + "/" + key
+                        } else {
+                            String::from(":") + ns + "/" + key
+                        },
+                        Deserialize::deserialize(e).unwrap(),
```
evaporei, line 252: `from_edn` and `unwrap`

src/deserialize/mod.rs
```diff
@@ -177,6 +186,128 @@ where
...
+impl<T> Deserialize for HashMap<String, T>
+where
+    T: Deserialize,
+{
+    fn deserialize(edn: &Edn) -> Result<Self, Error> {
+        match edn {
+            Edn::Map(_) => Ok(edn
+                .map_iter()
+                .ok_or_else(|| Error::Iter(format!("Could not create iter from {:?}", edn)))?
+                .map(|(key, e)| (key, Deserialize::deserialize(e).unwrap()))
```
evaporei, line 198: `from_edn` and `unwrap`

src/deserialize/mod.rs
```diff
@@ -177,6 +186,128 @@ where
...
+                .map_iter()
+                .ok_or_else(|| Error::Iter(format!("Could not create iter from {:?}", edn)))?
+                .map(|(key, e)| {
+                    (
+                        if ns.starts_with(":") {
+                            ns.to_string() + "/" + key
+                        } else {
+                            String::from(":") + ns + "/" + key
+                        },
+                        Deserialize::deserialize(e).unwrap(),
```
evaporei, line 213: `from_edn` and `unwrap`

src/deserialize/mod.rs
naomijub: every other deserialize uses `Deserialize::deserialize` and I think `from_edn`is something more usable from the client side, as well as `Deserialize::deserialize` is more explicit to have in an internal library context

naomijub: couldn't resolve `unwrap`

naomijub:
https://github.com/naomijub/edn-rs/issues/68

PR #73, opened 2020-10-21
Improves error messaging

naomijub:
Performance loss around 2.5 to 4% parsing example at 12us

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "edn-rs"
-version = "0.16.10"
+version = "0.16.11"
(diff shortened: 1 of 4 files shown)


naomijub:
This is a propose change for issue https://github.com/naomijub/edn-rs/issues/70

evaporei:
Most of it seems fine :+1: 
I just have two things to point out:

- (opinion) I think we could destructure instead of using `c.1`, and only use the index when necessary
- It doesn't seem to have a test case for `could not be parsed at char count` that checks the character index error

naomijub:
This commit adds tests 5d85693. However, destructuring seems to have impacted performance 🤔.
We can check this in another PR

PR #78, opened 2021-01-09
Update read_str

naomijub:
Update version and adds @dimfeld branch


naomijub:
Current master seems to solve the problem better, with no impact on performance and no extra deps

PR #93, opened 2022-05-15
Attempt to implement comments parsing for deserialize

jollm:
This change attempts to handle the following case:

"If a ; character is encountered outside of a string, that character
and all subsequent characters to the next newline should be ignored."

as described in https://github.com/edn-format/edn#comments

### DIFF
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -8,24 +8,28 @@ pub(crate) fn parse(
     c: Option<(usize, char)>,
     chars: &mut std::iter::Enumerate<std::str::Chars>,
 ) -> Result<Edn, Error> {
+    match parse_internal(c, chars)? {
+        Some(edn) => Ok(edn),
+        None => Ok(Edn::Empty),
+    }
+}
+
+fn parse_internal(
+    c: Option<(usize, char)>,
+    chars: &mut std::iter::Enumerate<std::str::Chars>,
+) -> Result<Option<Edn>, Error> {
     Ok(match c {
-        Some((_, '[')) => read_vec(chars)?,
-        Some((_, '(')) => read_list(chars)?,
-        Some((_, '#')) => tagged_or_set(chars)?,
-        Some((_, '{')) => read_map(chars)?,
+        Some((_, '[')) => Some(read_vec(chars)?),
+        Some((_, '(')) => Some(read_list(chars)?),
+        Some((_, '#')) => Some(tagged_or_set(chars)?),
+        Some((_, '{')) => Some(read_map(chars)?),
         Some((_, ';')) => {
-            let cmt = read_comment(chars)?;
-            match chars.next() {
-                Some(c) => parse(Some(c), chars)?,
-                None => cmt,
-            }
+            chars.skip_while(|c| c.1 != '\n').map(|c| c.1).next();
+            read_if_not_container_end(chars)?
         }
-        Some((_, s)) if s.is_whitespace() => match chars.next() {
-            Some(c) => parse(Some(c), chars)?,
-            None => Edn::Empty,
-        },
-        None => Edn::Empty,
-        edn => parse_edn(edn, chars)?,
+        Some((_, s)) if s.is_whitespace() => read_if_not_container_end(chars)?,
+        None => None,
+        edn => Some(parse_edn(edn, chars)?),
     })
 }
 
@@ -55,7 +59,7 @@ fn tagged_or_set(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Ed
 
 fn read_key_or_nsmap(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Edn, Error> {
     let mut key_chars = chars.clone().take_while(|c| {
-        !c.1.is_whitespace() && c.1 != ',' && c.1 != ')' && c.1 != ']' && c.1 != '}'
+        !c.1.is_whitespace() && c.1 != ',' && c.1 != ')' && c.1 != ']' && c.1 != '}' && c.1 != ';'
     });
     let c_len = key_chars.clone().count();
 
@@ -290,13 +294,12 @@ fn read_vec(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Edn, Er
     loop {
         match chars.next() {
             Some((_, ']')) => return Ok(Edn::Vector(Vector::new(res))),
-            Some((_, ';')) => {
-                read_comment(chars)?;
+            Some(c) if c.1 != ',' => {
+                if let Some(e) = parse_internal(Some(c), chars)? {
+                    res.push(e);
+                }
             }
-            Some(c) if !c.1.is_whitespace() && c.1 != ',' => {
-                res.push(parse(Some(c), chars)?);
-            }
-            Some(c) if c.1.is_whitespace() || c.1 == ',' => (),
+            Some(c) if c.1 == ',' => (),
             err => {
                 return Err(Error::ParseEdn(format!(
                     "{:?} could not be parsed at char count {}",
@@ -317,13 +320,12 @@ fn read_list(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Edn, E
     loop {
         match chars.next() {
             Some((_, ')')) => return Ok(Edn::List(List::new(res))),
-            Some((_, ';')) => {
-                read_comment(chars)?;
+            Some(c) if c.1 != ',' => {
+                if let Some(e) = parse_internal(Some(c), chars)? {
+                    res.push(e);
+                }
             }
-            Some(c) if !c.1.is_whitespace() && c.1 != ',' => {
-                res.push(parse(Some(c), chars)?);
-            }
-            Some(c) if c.1.is_whitespace() || c.1 == ',' => (),
+            Some(c) if c.1 == ',' => (),
             err => {
                 return Err(Error::ParseEdn(format!(
                     "{:?} could not be parsed at char count {}",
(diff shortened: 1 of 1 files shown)


src/deserialize/parse.rs
```diff
@@ -397,6 +426,10 @@ fn read_map(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Edn, Er
     loop {
         match chars.next() {
             Some((_, '}')) => return Ok(Edn::Map(Map::new(res))),
+            Some((_, ';')) => {
+                read_comment(chars)?;
+                ()
```
naomijub: You don't need `()` here (Rust will interpret a line ending in `;` as `()`.

src/deserialize/parse.rs
```diff
@@ -13,6 +13,14 @@ pub(crate) fn parse(
         Some((_, '(')) => read_list(chars)?,
         Some((_, '#')) => tagged_or_set(chars)?,
         Some((_, '{')) => read_map(chars)?,
+        Some((_, ';')) => {
+            let cmt = read_comment(chars)?;
+            match chars.next() {
```
naomijub: Considering read_comment returns `Edn::Empty` and the functions receives a mutable chars, I don't think this match is necessary

src/deserialize/parse.rs
```diff
@@ -26,8 +34,13 @@ pub(crate) fn parse_edn(
         Some((_, ':')) => read_key_or_nsmap(chars),
         Some((_, '-')) => Ok(read_number('-', chars)?),
         Some((_, '\\')) => Ok(read_char(chars)?),
+        Some((_, ';')) => {
+            read_comment(chars)?;
+            Ok(parse_edn(chars.next(), chars)?)
```
naomijub: if you call `parse_edn` here you will be missing the full parse (`parse`), as `parse_edn` only parses "primitive" types. Also, I am not sure you actually need this line `Ok(parse_edn(chars.next(), chars)?)`, maybe only if comments come as the first line of code.

src/deserialize/parse.rs
```diff
@@ -13,6 +13,14 @@ pub(crate) fn parse(
...
         Some((_, '#')) => tagged_or_set(chars)?,
         Some((_, '{')) => read_map(chars)?,
+        Some((_, ';')) => {
+            let cmt = read_comment(chars)?;
+            match chars.next() {
+                Some(c) => parse(Some(c), chars)?,
+                None => cmt,
+            }
+        }
+        Some((_, s)) if s.is_whitespace() => parse(chars.next(), chars)?,
```
naomijub: I don't get why we need this extra parse? the code in general already discards whitespaces and commas.

src/deserialize/parse.rs
```diff
@@ -328,6 +349,10 @@ fn read_set(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Edn, Er
     loop {
         match chars.next() {
             Some((_, '}')) => return Ok(Edn::Set(Set::new(res))),
+            Some((_, ';')) => {
+                read_comment(chars)?;
+                ()
```
naomijub: You don't need () here (Rust will interpret a line ending in ; as ().

src/deserialize/parse.rs
```diff
@@ -360,6 +385,10 @@ fn read_namespaced_map(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Res
     loop {
         match chars.next() {
             Some((_, '}')) => return Ok(Edn::NamespacedMap(namespace, Map::new(res))),
+            Some((_, ';')) => {
+                read_comment(chars)?;
+                ()
```
naomijub: You don't need () here (Rust will interpret a line ending in ; as ().

src/deserialize/parse.rs
naomijub: I do think this expression is not necessary, as it is already being considered in `parse`

src/deserialize/parse.rs
```diff
@@ -421,6 +454,17 @@ fn read_map(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Edn, Er
     }
 }
 
+fn read_comment(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Edn, Error> {
+    loop {
```
naomijub: we usually use `chars.take_while(|c| c != \n)`, instead of loops, as it is lazy. Also, have you thought about a test case when the comment doesnt end in a `\n`? I think there is a specific value that you can pass that encompasses new lines, eof, and returns.

src/deserialize/parse.rs
```diff
@@ -540,6 +604,25 @@ mod test {
         );
     }
 
+    #[test]
+    fn parse_comment_in_simple_vec() {
+        let mut edn = "[11 \"2\" 3.3 ; float in simple vec\n:b true \\c]"
```
naomijub, line 626: I would have some tests that dont actually use `\n` and use strings like:
```
let mut edn = "[
  11 
  \"2\" 
  3.3 
 ;; float in simple vec
 :b 
true 
\\c
]"
```

```

**Review by naomijub — CHANGES_REQUESTED** [review]
Nice works
Thank you for the PR.

File: src/deserialize/parse.rs (hunk 1 above)
Comment (jollm) [reply src/deserialize/parse.rs]: thanks for the tip on that!, I'll fix all of those similar cases

the top level parse doesn't catch inside a map, vec, set, etc., so if i comment this line the following test fails (from line 852):
```test deserialize::parse::test::parse_comment_inside_map ... FAILED```

File: src/deserialize/parse.rs (hunk 3 above)
Comment (jollm) [reply src/deserialize/parse.rs]: only parse calls parse_edn, so if parse_edn calls itself, it will still return back to parse?

Yes that is for comments as the first line, e.g. test parse_str_top_level_comment on line 494

File: src/deserialize/parse.rs (hunk 4 above)
Comment (jollm) [reply src/deserialize/parse.rs]: independent of the case with comments, if i comment that line, e.g. the following test will fail:
```
    #[test]
    fn parse_simple_map() {
        let mut edn = " {:a \"2\" :b false :c nil }".chars().enumerate();

        assert_eq!(
            Edn::Map(Map::new(
                map! {":a".to_string() => Edn::Str("2".to_string()),
                ":b".to_string() => Edn::Bool(false), ":c".to_string() => Edn::Nil}
            ))
        );
    }
```
because of not handling whitespace at the top level (unless I'm missing where this is done at another level, which is entirely possible)

src/deserialize/parse.rs
jollm: it handles EOF by matching None

I've considered files that use \r instead of \n, probably good to handle that as the spec doesn't emphatically say only \n line endings..

will also change to lazy version

src/deserialize/parse.rs
jollm: If I copy that in verbatim it passes. Is that what you intended?

naomijub:
@evaporei Do you think we need a type `Edn::Comment` for serialization purposes?

src/deserialize/parse.rs
jollm: This is for the case where the only thing existing is a comment (tests starting on line 815) (fails without match). Normally the return from read_comment is ignored in favor of whatever the next call to parse returns. Your comment made me realize that an attempt to deserialize an empty or whitespace only string will fail instead of return Edn::Empty, so it would be surprising for comment to behave differently. For now I will make both return Edn::Empty, but the other option is to have both fail..

src/deserialize/parse.rs
naomijub: `parse`  calls `parse_edn`, https://github.com/naomijub/edn-rs/blob/master/src/deserialize/parse.rs#L16.
I think this is the reason why 
> the top level parse doesn't catch inside a map, vec, set, etc., so if i comment this line the following test fails (from line 852): `test deserialize::parse::test::parse_comment_inside_map ... FAILED`

is failing

src/deserialize/parse.rs
naomijub: Yes

naomijub:
@jollm I noticed the circleCI hasn't executed. You need to go to your fork circle CI and enable it for this project

src/deserialize/parse.rs
jollm: probably "and all subsequent characters to the next newline should be ignored" is meant to imply newline endings, otherwise it would need to be known whether to expect \r or \n or \r\n?

src/deserialize/parse.rs
jollm: It calls back to top level now.

src/deserialize/parse.rs
jollm: Changed to lazy iterator expression.

src/deserialize/parse.rs
jollm: Added a test like this

src/deserialize/parse.rs
jollm: I see what you mean, this shouldn't be needed, I have removed these from parse_edn

jollm:
The latest diff is smaller and I hope cleaner. It avoids checking for comments and whitespace at inner levels.

naomijub:
I promise to review again tomorrow

naomijub:
Dont forget to enable your circle CI for this project https://app.circleci.com/pipelines/github/jollm/edn-rs

jollm:
> Dont forget to enable your circle CI for this project https://app.circleci.com/pipelines/github/jollm/edn-rs

https://app.circleci.com/pipelines/github/jollm/edn-rs/1/workflows/5d73314a-58b5-45f1-9729-5664ca252586

naomijub (approved):

PR #98, opened 2022-05-26
Try to implement discard sequence #_ for deserialize

jollm:
See https://github.com/naomijub/edn-rs/issues/94

Noticed your issue post after PR #93
Hope it's okay I tried to implement this one, seemed maybe fun to try

### DIFF
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -1017,6 +1017,23 @@ mod test {
     }
 
     #[test]
+    fn parse_discard_inside_seq() {
+        let mut edn = "#_\"random comment\" [:a :b :c #_(:hello :world) :d]"
+            .chars()
+            .enumerate();
+        let res = parse(edn.next(), &mut edn).unwrap();
+        assert_eq!(
+            res,
+            Edn::Vector(Vector::new(vec![
+                Edn::Key(":a".to_string()),
+                Edn::Key(":b".to_string()),
+                Edn::Key(":c".to_string()),
+                Edn::Key(":d".to_string())
+            ]))
+        );
+    }
+
+    #[test]
     fn parse_map_keyword_with_commas() {
         let mut edn = "{ :a :something, :b false, :c nil, }".chars().enumerate();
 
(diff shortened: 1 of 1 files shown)


naomijub:
Loved it! Will check as soon as I am free

src/deserialize/parse.rs
```diff
@@ -907,6 +926,96 @@ mod test {
...
+            .chars()
+            .enumerate();
+        let res = parse(edn.next(), &mut edn).unwrap();
+
+        assert_eq!(res, Edn::Empty)
+    }
+
+    #[test]
+    fn parse_discard_repeat_not_empty() {
+        let mut edn = "#_ ,, #_{discard again} #_ {:and :again} :okay {:a map}"
```
naomijub, line 993: this seems weird, parsing `:okay {:a map}` seems like a problem to return:
```
Edn::Map(Map::new(
  map! {":a".to_string() => Edn::Symbol("map".to_string())}
))
```

Did you check the standard response? I would expect the lib to return the keyword :okay and ignore the map

src/deserialize/parse.rs
```diff
@@ -907,6 +926,96 @@ mod test {
         )
     }
 
+    #[test]
```
naomijub, line 929: I would say it is missing a test where the discard element is a sequence inside a sequence, like:
```
#_"random comment" [:a :b :c #_(:hello :world) :d]
``` 
then the result would be a vector with `[:a :b :c :d]`

src/deserialize/parse.rs
jollm: yea let me check what standard lib does with that.

this is how it works here: 
```
(discard 
  (discard {discard again}
           return (discard {:and :again}
                           return :okay)))
  return {:a map})
```

jollm: Here it is:
```test.clj
clj-scratch.core> (edn/read-string "#_ ,, #_{discard again} #_ {:and :again} :okay {:a map}")
{:a map}
clj-scratch.core> 
```

naomijub (approved):

src/deserialize/parse.rs
naomijub: Nice

jollm:
Wait, i found one that breaks:
```test.clj
clj-scratch.core> (edn/read-string "[:foo #_ foo]")
[:foo]
clj-scratch.core> 
```

Have a fix ready..

naomijub:
I think you will need to rebase or create a new PR

Issue #100, opened 2023-08-25
Handling of non-dec numbers (eg "0x2a")


Grinkers:
According to the spec, this isn't an issue. Here's a very relavent discussion I found while looking into this.
https://github.com/edn-format/edn/issues/47

here is some examples of read-string
```
(clojure.edn/read-string "0x2a") -> 42
(clojure.edn/read-string "2r101010") -> 42
(clojure.edn/read-string "36r16") -> 42
```

As far as this crate is concerned, I think maybe it should be more liberal than the spec for some convenient patterns. In particular, "0x" being supported in both rust and clojure.edn/read-string. "2r" vs "0b" is a bit harder of a thing and could be a future nightmare. Alternatively, perhaps this crate should be more apparent with off-spec integers so these doesn't succeed.

```
assert_eq!(Edn::from_str("0x2a").unwrap().to_int().unwrap(), 0);
assert_eq!(Edn::from_str("42FOOBAR9000").unwrap().to_int().unwrap(), 42);
```

naomijub:
Seems interesting, are you able to open a PR for this?

PR #101, opened 2023-08-26
Fix an issue with parsing false/true/nil when no ' ' character is following.

Grinkers:
true, false, and nil were being parsed as a Symbol if there was no space or delimiter immediately following. Here's the output of the new tests before the fix:
```
failures:

---- deserialize::parse::test::parse_bool_in_crlf_newline_simple_vec_str_literal stdout ----
thread 'deserialize::parse::test::parse_bool_in_crlf_newline_simple_vec_str_literal' panicked at 'assertion failed: `(left == right)`
  left: `Vector(Vector([Symbol("nil"), Symbol("false"), Symbol("true")]))`,
 right: `Vector(Vector([Nil, Bool(false), Bool(true)]))`', src/deserialize/parse.rs:753:9

---- deserialize::parse::test::parse_bool_in_tab_simple_vec_str_literal stdout ----
thread 'deserialize::parse::test::parse_bool_in_tab_simple_vec_str_literal' panicked at 'assertion failed: `(left == right)`
  left: `Vector(Vector([Symbol("true"), Symbol("nil"), Symbol("false")]))`,
 right: `Vector(Vector([Bool(true), Nil, Bool(false)]))`', src/deserialize/parse.rs:741:9

---- deserialize::parse::test::parse_bool_in_newline_simple_vec_str_literal stdout ----
thread 'deserialize::parse::test::parse_bool_in_newline_simple_vec_str_literal' panicked at 'assertion failed: `(left == right)`
  left: `Vector(Vector([Symbol("true"), Symbol("false"), Symbol("nil")]))`,
 right: `Vector(Vector([Bool(true), Bool(false), Nil]))`', src/deserialize/parse.rs:729:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    deserialize::parse::test::parse_bool_in_crlf_newline_simple_vec_str_literal
    deserialize::parse::test::parse_bool_in_newline_simple_vec_str_literal
    deserialize::parse::test::parse_bool_in_tab_simple_vec_str_literal
```

I accidentally found this when my editor was stripping whitespace here https://github.com/naomijub/edn-rs/blob/e74e564afb25a8011fa8437d35bf5e66b5a2eb92/src/deserialize/parse.rs#L704 while navigating about the code base thinking about #100.

On a related note, I'm not able to see anything on CircleCI. There's a few clippy warnings and a lot of formatting things, at least on rust 1.71 using the default rustfmt (1.5.2). Any thoughts on this? Maybe github workflows?

### DIFF
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -1,6 +1,8 @@
 use crate::edn::{Edn, Error, List, Map, Set, Vector};
 use std::collections::{BTreeMap, BTreeSet};
 
+const DELIMITERS: [char; 5] = [',', ']', '}', ')', ';'];
+
 pub(crate) fn tokenize(edn: &str) -> std::iter::Enumerate<std::str::Chars> {
     edn.chars().enumerate()
 }
@@ -262,12 +264,11 @@ fn read_bool_or_nil(
         .next()
         .ok_or_else(|| Error::ParseEdn("Could not identify symbol index".to_string()))?
         .0;
-    let delimiters = [',', ']', '}', ')', ';'];
     match c {
         't' if {
             let val = chars
                 .clone()
-                .take_while(|(_, c)| !c.is_whitespace() && !delimiters.contains(c))
+                .take_while(|(_, c)| !c.is_whitespace() && !DELIMITERS.contains(c))
                 .map(|c| c.1)
                 .collect::<String>();
             val.eq("rue")
@@ -282,7 +283,7 @@ fn read_bool_or_nil(
         'f' if {
             let val = chars
                 .clone()
-                .take_while(|(_, c)| !c.is_whitespace() && !delimiters.contains(c))
+                .take_while(|(_, c)| !c.is_whitespace() && !DELIMITERS.contains(c))
                 .map(|c| c.1)
                 .collect::<String>();
             val.eq("alse")
@@ -297,7 +298,7 @@ fn read_bool_or_nil(
         'n' if {
             let val = chars
                 .clone()
-                .take_while(|(_, c)| !c.is_whitespace() && !delimiters.contains(c))
+                .take_while(|(_, c)| !c.is_whitespace() && !DELIMITERS.contains(c))
                 .map(|c| c.1)
                 .collect::<String>();
             val.eq("il")
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -262,7 +262,7 @@ fn read_bool_or_nil(
         .next()
         .ok_or_else(|| Error::ParseEdn("Could not identify symbol index".to_string()))?
         .0;
-    let delimiters = [',', ']', '}', ')'];
+    let delimiters = [',', ']', '}', ')', ';'];
     match c {
         't' if {
             let val = chars
@@ -845,6 +845,18 @@ mod test {
                 Edn::Char('c'),
                 Edn::UInt(3)
             ]))
+        )
+    }
+
+    #[test]
+    fn parse_true_false_nil_with_comments_in_set() {
+        let mut edn = "#{true;this is true\nfalse;this is false\nnil;this is nil\n}"
+            .chars()
+            .enumerate();
+
+        assert_eq!(
+            parse(edn.next(), &mut edn).unwrap(),
+            Edn::Set(Set::new(set![Edn::Bool(true), Edn::Bool(false), Edn::Nil,]))
         )
     }
 
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -262,15 +262,15 @@ fn read_bool_or_nil(
         .next()
         .ok_or_else(|| Error::ParseEdn("Could not identify symbol index".to_string()))?
         .0;
+    let delimiters = [',', ']', '}', ')'];
     match c {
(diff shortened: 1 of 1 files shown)


Grinkers:
force-pushed after running rustfmt on these changes...

Grinkers:
Fixes an additional edge case for valid edn
```
#{true;this is true
false;this is false
nil;this is nil
}
```

```
---- deserialize::parse::test::parse_true_false_nil_with_comments_in_set stdout ----
thread 'deserialize::parse::test::parse_true_false_nil_with_comments_in_set' panicked at 'assertion failed: `(left == right)`
  left: `Set(Set({Symbol("false;this"), Symbol("is"), Symbol("nil;this"), Symbol("true;this"), Bool(false), Bool(true), Nil}))`,
 right: `Set(Set({Bool(false), Bool(true), Nil}))`', src/deserialize/parse.rs:857:9
 ```
 
 result of test before the additional ';' delimiter.

File: src/deserialize/parse.rs [hunk 1]
```diff
@@ -262,15 +262,15 @@ fn read_bool_or_nil(
         .next()
         .ok_or_else(|| Error::ParseEdn("Could not identify symbol index".to_string()))?
         .0;
+    let delimiters = [',', ']', '}', ')', ';'];
```
naomijub: constify

src/deserialize/parse.rs
```diff
@@ -262,15 +262,15 @@ fn read_bool_or_nil(
...
-            let val = chars.clone().take(4).map(|c| c.1).collect::<String>();
-            val.eq("rue ")
-                || val.eq("rue,")
-                || val.eq("rue]")
-                || val.eq("rue}")
-                || val.eq("rue)")
-                || val.eq("rue")
+            let val = chars
+                .clone()
+                .take_while(|(_, c)| !c.is_whitespace() && !delimiters.contains(c))
```
naomijub: I think the complete linter will ask you to replace this not AND not logic

naomijub (changes requested):
Just that one constant change. Please address the clippy one before this one

src/deserialize/parse.rs
Grinkers: Unless I'm missing something
both 
`const delimiters: [char; 5] = [',', ']', '}', ')', ';'];`
and 
`let delimiters = [',', ']', '}', ')', ';'];`
are [char; 5] and don't require an allocation. I find the latter easier to read and less manual type typing. Open to suggestions

src/deserialize/parse.rs
Grinkers: > 

I will rebase after #102 so we can let tooling audit.

Grinkers: I don't see anything for 
cargo clippy -- -W clippy::all -W clippy::pedantic -W clippy::nursery --deny warnings
suggestions?

src/deserialize/parse.rs
naomijub: const outside the function call, as a global const:

Global: 
`const DELIMITERS: [CHAR;5] = [',', ']', '}', ')', ';'];`

I don't think I have ever used const inside a function, but yeah, they would be +- the same thing

naomijub:
Just this one change, I do think it would improve the code usage

Grinkers:
Changed to a const global for easier reuse, which is a good reason to do this (I can rebase non-dec to reuse after merging).

However from a performance perspective, I believe this will actually be worse. LLVM is crazy good at unraveling locally scoped things like this. No allocations and everything is optimized for avoiding cache misses.
https://godbolt.org/z/5q8MvGhY4
There's no point in playing performance over something so small while there's clones.

naomijub:
Do you mind using git ref in cargo while we dont fix the clippy issues?

naomijub:
> Changed to a const global for easier reuse, which is a good reason to do this (I can rebase non-dec to reuse after merging).
> 
> However from a performance perspective, I believe this will actually be worse. LLVM is crazy good at unraveling locally scoped things like this. No allocations and everything is optimized for avoiding cache misses. https://godbolt.org/z/5q8MvGhY4 There's no point in playing performance over something so small while there's clones.

I did not expect this, but I guess, in this case the performance hit is small enough to consider the usability as a bonus. Thanks for the lesson

Grinkers:
> Do you mind using git ref in cargo while we dont fix the clippy issues?

In cargo? What do you mean and for what? Hopefully you're fine with reordering I just did in #103 before I saw anything. I think this is closed.

naomijub:
I mean for developing using the library, something like this https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories. 

I would like to address the clippy suggestions before releasing a new version

Grinkers:
Ah, I see. No worries. I actually haven't picked up this crate up yet, but I probably will this week. I need non-dec so I was planning on just forking and pointing to that in the worst case (following clojure.edn/read-string instead of the EDN spec).

Btw thanks for the project, I've found it very easy to navigate. I don't need futures or json, so not having any additional dependencies is fantastic.

naomijub:
This project is actually used in a clojure project for iterating large amounts of data, so not having dependencies was crucial. Happy that you liked it

PR #102, opened 2023-08-26
Add Github Workflows + clippy fixes

Grinkers:
Workflows before and after clippy fixes. No functionality changes are intended.

before:
https://github.com/Grinkers/edn-rs/actions/runs/5983591578

after:
https://github.com/Grinkers/edn-rs/actions/runs/5983672308

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,52 @@
+name: Check
+
+on:
+  push:
+    branches:
+  pull_request:
+    branches:
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: build
+        run: cargo build --verbose
+
+  cargo_test:
+    name: cargo test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: example_fast
+        run: cargo test --examples --no-fail-fast
+      - name: example_json_to_edn
+        run: cargo test --example json_to_edn --features "json"
+      - name: example_edn_to_json
+        run: cargo test --example edn_to_json --features "json"
+      - name: example_async
+        run: cargo run --example async --features "async"
+
+  fmt:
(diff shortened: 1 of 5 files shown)


.github/workflows/check.yml
```diff
@@ -0,0 +1,52 @@
+name: Check
```
naomijub: nice, this has been a pending issue for a while

.github/workflows/check.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  push:
+    branches:
+  pull_request:
+    branches:
```
naomijub: please use:
```rs
name: Rust

on:
  push:
    branches: [ "main" ]
  pull_request:
    branches: [ "*" ]
```

.github/workflows/check.yml
naomijub: Standard seems to call it Rust.yml, but this is not an issue

.github/workflows/check.yml
```diff
@@ -0,0 +1,52 @@
...
+
+  clippy_check:
+      name: clippy
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v3
+        - uses: dtolnay/rust-toolchain@stable
+          with:
+            components: clippy
+        - run: cargo clippy --all-features -- --deny warnings
```
naomijub: We also use pedantic and nursery:
        - run: cargo clippy --all-features -- -W clippy::all -W clippy::pedantic -W clippy::nursery --deny warnings

naomijub (changes requested):
Nice, just a couple of changes

Grinkers:
btw with pedantic and nursey, there's many pages of output. Most of them seem decent, but some are what I consider false positives like
https://rust-lang.github.io/rust-clippy/master/index.html#/must_use_candidate

I'm willing to help maintain and go through everything if #103 can get through. Without clojure.edn/read-string behavior, I can't consider using this crate at this time.

naomijub:
I think I can try to address the remaining issues this week. Thanks

PR #103, opened 2023-08-27
Implementation of non-dec number parsing, trying to match clojure.edn/read-string as much as possible.

### LINKED ISSUE #100: Handling of non-dec numbers (eg "0x2a")

Grinkers:
Resolves #100, both the mentioned non-dec parsing but also throwing errors when encountering invalid numbers instead of giving a bad result.

Using rust's `from_str_radix`, supporting all radix from 2 to 36, just like `clojure.edn/read-string` ended up not being too painful. 

I don't have a super strong opinion other than needing 0x support before picking up this crate for my projects. That being said, I can only imagine future problems/compatibility by adding rust's additional `0b` and `0o` patterns.

### DIFF
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -206,27 +206,71 @@ fn read_number(n: char, chars: &mut std::iter::Enumerate<std::str::Chars>) -> Re
         .0;
     let c_len = chars
         .clone()
-        .take_while(|c| {
-            c.1.is_numeric()
-                || c.1 == '.'
-                || c.1 == '/'
-                || c.1 == 'e'
-                || c.1 == 'E'
-                || c.1 == '+'
-                || c.1 == '-'
-        })
+        .take_while(|(_, c)| !c.is_whitespace() && !DELIMITERS.contains(c))
         .count();
-    let mut number = String::new();
-    let string = chars.take(c_len).map(|c| c.1).collect::<String>();
-    number.push(n);
-    number.push_str(&string);
+    let (number, radix) = {
+        let mut number = String::new();
+        number.push(n);
+        for (_, c) in chars.take(c_len) {
+            number.push(c)
+        }
+        if number.to_lowercase().starts_with("0x") {
+            number.remove(0);
+            number.remove(0);
+            (number, 16)
+        } else if number.to_lowercase().starts_with("-0x") {
+            number.remove(1);
+            number.remove(1);
+            (number, 16)
+        } else if let Some(index) = number.to_lowercase().find('r') {
+            let negative = number.starts_with('-');
+            let radix = {
+                if negative {
+                    (number[1..index]).parse::<u32>()
+                } else {
+                    (number[0..index]).parse::<u32>()
+                }
+            };
+            match radix {
+                Ok(r) => {
+                    // from_str_radix panics if radix is not in the range from 2 to 36
+                    if r > 36 || r < 2 {
+                        return Err(Error::ParseEdn(format!("Radix of {} is out of bounds", r)));
+                    }
+
+                    if negative {
+                        for _ in 0..(index) {
+                            number.remove(1);
+                        }
+                    } else {
+                        for _ in 0..(index + 1) {
+                            number.remove(0);
+                        }
+                    }
+                    (number, r)
+                }
+                Err(e) => {
+                    return Err(Error::ParseEdn(format!(
+                        "{} while trying to parse radix from {}",
+                        e, number
+                    )));
+                }
+            }
+        } else {
+            (number, 10)
+        }
+    };
 
     match number {
         n if (n.contains('E') || n.contains('e')) && n.parse::<f64>().is_ok() => {
             Ok(Edn::Double(n.parse::<f64>()?.into()))
         }
-        n if n.parse::<usize>().is_ok() => Ok(Edn::UInt(n.parse::<usize>()?)),
-        n if n.parse::<isize>().is_ok() => Ok(Edn::Int(n.parse::<isize>()?)),
+        n if usize::from_str_radix(&n, radix).is_ok() => {
+            Ok(Edn::UInt(usize::from_str_radix(&n, radix)?))
+        }
+        n if isize::from_str_radix(&n, radix).is_ok() => {
+            Ok(Edn::Int(isize::from_str_radix(&n, radix)?))
+        }
         n if n.parse::<f64>().is_ok() => Ok(Edn::Double(n.parse::<f64>()?.into())),
         n if n.contains('/') && n.split('/').all(|d| d.parse::<f64>().is_ok()) => {
             Ok(Edn::Rational(n))
@@ -236,8 +280,8 @@ fn read_number(n: char, chars: &mut std::iter::Enumerate<std::str::Chars>) -> Re
             read_symbol(n.next().unwrap_or(' '), &mut n.enumerate())
         }
         _ => Err(Error::ParseEdn(format!(
-            "{} could not be parsed at char count {}",
-            number, i
+            "{} could not be parsed at char count {} with radix {}",
+            number, i, radix
         ))),
     }
 }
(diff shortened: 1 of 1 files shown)


src/deserialize/parse.rs
```diff
@@ -197,34 +197,79 @@ fn read_discard(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Opt
 }
 
 fn read_number(n: char, chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Edn, Error> {
+    let delimiters = [',', ']', '}', ')', ';'];
```
naomijub: This could be a constant, it would prevent reallocatiion for every time this function is called.

src/deserialize/parse.rs
```diff
@@ -197,34 +197,79 @@ fn read_discard(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Opt
...
-    number.push_str(&string);
+    let (number, radix) = {
+        let mut number = String::new();
+        number.push(n);
+        for (_, c) in chars.take(c_len) {
+            number.push(c)
+        }
+        if number.to_lowercase().starts_with("0x") {
+            number.remove(0);
+            number.remove(0);
```
naomijub, line 219: Why not use &str? you could substring it with `number[2..]`, instead of removing twice

src/deserialize/parse.rs
```diff
@@ -197,34 +197,79 @@ fn read_discard(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Opt
...
+            (number, 16)
+        } else if number.to_lowercase().starts_with("-0x") {
+            number.remove(1);
+            number.remove(1);
```
naomijub, line 223: same here

src/deserialize/parse.rs
```diff
@@ -197,34 +197,79 @@ fn read_discard(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Opt
...
+                        return Err(Error::ParseEdn(format!("Radix of {} is out of bounds", r)));
+                    }
+
+                    if negative {
+                        for _ in 0..(index) {
+                            number.remove(1);
+                        }
+                    } else {
+                        for _ in 0..(index + 1) {
+                            number.remove(0);
```
naomijub, line 247: I'm not a fan of removing content from strings

src/deserialize/parse.rs
```diff
@@ -197,34 +197,79 @@ fn read_discard(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Opt
...
+            let radix = {
+                if negative {
+                    (number[1..index]).parse::<u32>()
+                } else {
+                    (number[0..index]).parse::<u32>()
+                }
+            };
+            match radix {
+                Ok(r) => {
+                    // from_str_radix panics if radix is not in the range from 2 to 36
```
naomijub, line 236: Noted

src/deserialize/parse.rs
```diff
@@ -1424,4 +1469,72 @@ mod test {
...
+        assert_eq!(res, Edn::UInt(42));
+
+        let mut edn = "-0X2A".chars().enumerate();
+        let res = parse(edn.next(), &mut edn).unwrap();
+        assert_eq!(res, Edn::Int(-42));
+    }
+
+    #[test]
+    fn parse_radix_ints() {
+        let mut edn = "16r2a".chars().enumerate();
```
naomijub, line 1538: I would expect this to be a symbol

src/deserialize/parse.rs
```diff
@@ -1424,4 +1469,72 @@ mod test {
...
+        let res = parse(edn.next(), &mut edn).unwrap();
+        assert_eq!(res, Edn::UInt(42));
+
+        let mut edn = "8r63".chars().enumerate();
+        let res = parse(edn.next(), &mut edn).unwrap();
+        assert_eq!(res, Edn::UInt(51));
```
naomijub, line 1492: Shouldn't this panic?

src/deserialize/parse.rs
```diff
@@ -1424,4 +1469,72 @@ mod test {
...
+
+        let mut edn = "36rabcxyz".chars().enumerate();
+        let res = parse(edn.next(), &mut edn).unwrap();
+        assert_eq!(res, Edn::UInt(623741435));
+
+        let mut edn = "-16r2a".chars().enumerate();
```
naomijub, line 1550: @evaporei need you opinion here, I don't like the idea of having a negative sign before the radix, like `-16r`, and I would be keen to using something like `-128426_r9`, what do you think? 

I also know it break style with the `0x`and `0b`, but I found this really hard to read. Another hard example would be `"36rabcxyz"`. I would expect this to be a symbol

src/deserialize/parse.rs
```diff
@@ -1424,4 +1469,72 @@ mod test {
...
+        assert_eq!(res, Edn::Int(-42));
+
+        let mut edn = "-32rFOObar".chars().enumerate();
+        let res = parse(edn.next(), &mut edn).unwrap();
+        assert_eq!(res, Edn::Int(-529280347));
+    }
+
+    #[test]
+    fn parse_invalid_ints() {
+        let mut edn = "42invalid123".chars().enumerate();
```
naomijub, line 1561: isn't this a valid symbol?

src/deserialize/parse.rs
```diff
@@ -1424,4 +1469,72 @@ mod test {
...
+        assert_eq!(
+            parse(edn.next(), &mut edn),
+            Err(Error::ParseEdn(
+                "42invalid123 could not be parsed at char count 1 with radix 10".to_string()
+            ))
+        );
+
+        let mut edn = "0xxyz123".chars().enumerate();
```
naomijub, line 1569: isn't this a valid symbol?

src/deserialize/parse.rs
```diff
@@ -1424,4 +1469,72 @@ mod test {
...
+        assert_eq!(
+            parse(edn.next(), &mut edn),
+            Err(Error::ParseEdn(
+                "xyz123 could not be parsed at char count 1 with radix 16".to_string()
+            ))
+        );
+
+        let mut edn = "42rabcxzy".chars().enumerate();
```
naomijub, line 1577: isn't this a valid symbol?

src/deserialize/parse.rs
```diff
@@ -1424,4 +1469,72 @@ mod test {
...
+        assert_eq!(
+            parse(edn.next(), &mut edn),
+            Err(Error::ParseEdn("Radix of 42 is out of bounds".to_string()))
+        );
+
+        let mut edn = "42crazyrabcxzy".chars().enumerate();
```
naomijub, line 1583: isn't this a valid symbol?

src/deserialize/parse.rs
```diff
@@ -197,34 +197,79 @@ fn read_discard(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Opt
...
-    let string = chars.take(c_len).map(|c| c.1).collect::<String>();
-    number.push(n);
-    number.push_str(&string);
+    let (number, radix) = {
+        let mut number = String::new();
+        number.push(n);
+        for (_, c) in chars.take(c_len) {
+            number.push(c)
+        }
+        if number.to_lowercase().starts_with("0x") {
```
naomijub, line 217: I would create a new parser for stuff staring with `0x` and not reuse this one

src/deserialize/parse.rs
```diff
@@ -197,34 +197,79 @@ fn read_discard(chars: &mut std::iter::Enumerate<std::str::Chars>) -> Result<Opt
...
+            number.remove(0);
+            number.remove(0);
+            (number, 16)
+        } else if number.to_lowercase().starts_with("-0x") {
+            number.remove(1);
+            number.remove(1);
+            (number, 16)
+        } else if let Some(index) = number.to_lowercase().find('r') {
```
naomijub, line 225: like I sair previously, I'm keen on having the radix at the very end, as `number_r8`, easier to parse as well. As this is something completely new in edn, I would go with this

src/deserialize/parse.rs
Grinkers: repeat of  https://github.com/naomijub/edn-rs/pull/101#discussion_r1306860246
will revisit after that's resolved, as that'll probably be first.

Grinkers:
Instead of commenting on a lot of the repeated areas, I was trying to match clojure.edn/read-string as much as possible. Here's some input/outputs

```
(clojure.edn/read-string "2r10100101")
165

(clojure.edn/read-string "16r2a")
42

(clojure.edn/read-string "16r2a")
42

(clojure.edn/read-string "-16r2a")
-42

(clojure.edn/read-string "-0X2A")
-42

(clojure.edn/read-string "-36rabcxyz")
-623741435

(clojure.edn/read-string "42invalid123")
java.lang.NumberFormatException: Invalid number: 42invalid123 [at <repl>:5:1]

(clojure.edn/read-string "0xxyz123")
java.lang.NumberFormatException: Invalid number: 0xxyz123 [at <repl>:1:1]

(clojure.edn/read-string "42rabcxzy")
java.lang.NumberFormatException: Radix out of range [at <repl>:2:1]

(clojure.edn/read-string "42crazyrabcxzy")
java.lang.NumberFormatException: Invalid number: 42crazyrabcxzy [at <repl>:3:1]
```

https://github.com/edn-format/edn#symbols

> Symbols begin with a non-numeric character

I realize a lot of this radix stuff is pretty messy (both readability and for parsing) and is technically off-spec, so I would understand if you don't want to take on the burden. I've become dependent (perhaps wrongly so!) on the more liberal behavior of clojure.edn/read-string, so I tried to match it as close as possible. This certainly seems like a pain point for EDN.

src/deserialize/parse.rs
Grinkers: The awful removes are due to it hard creating a slice `-a55a` from something like -16ra55a. This could be made more human readable, but I think it would require creating another allocation with another String

src/deserialize/parse.rs
naomijub: as responded later, this is the expected behavior in clojure

src/deserialize/parse.rs
naomijub: actually, a rebase would fix this

src/deserialize/parse.rs
Grinkers: I think this should live here, but am open to suggestions. 

I think moving chars (all about parse.rs) as a &str would allow for more powerful pattern matching without having to mutate the iterator. For example it might be nice to be able to match against `-0x...` and `0x...` here
https://github.com/naomijub/edn-rs/blob/8353fc4cbfb1938849ffed0a4a53951b95706f24/src/deserialize/parse.rs#L41

This would also likely be a huge performance gain, because there wouldn't need to be heap allocating clones just for reading the index or lengths of things, just to be tossed away.

I think this goes beyond the scope of non-dec and should be done independently (and probably after #104)

naomijub: Also thought about adding it to here `pub(crate) fn parse_edn`

Grinkers: I don't think this should be cleaned up without #105.

I'm biased because I specifically need non-doc, but also I think it might end up being here anyway. The pattern of needing to escape out 0x Nr, with minus symbols, will need an allocation and should probably be together anyway. I think #105 will make things obvious and at the very least a working implementation with test coverage should happen first.

naomijub:
Sorry for the delay, crazy times. I agree with you point, will merge

PR #108, opened 2023-09-11
Properly handle lisp quotes (' symbol) and numbers starting with '+'

### LINKED ISSUE #105: Can't parse symbols according to spec. More powerful "parse_edn" pattern matching?
According to https://github.com/edn-format/edn#symbols the following is valid (and confirmed with clojure.edn/read-string).

```
(clojure.edn/read-string "-foobar")
-foobar
(type (clojure.edn/read-string "-foobar"))
clojure.lang.Symbol
```

Currently parse_edn is only matching on the first character, which is insufficient for this case. `-foobar` gets treated as a Number.
https://github.com/naomijub/edn-rs/blob/8353fc4cbfb1938849ffed0a4a53951b95706f24/src/deserialize/parse.rs#L41

I believe it should be possible to be passing around a `&str` instead of `&mut std::iter::Enumerate<std::str::Chars>`, which would allow for str pattern matching. I don't know if this is a slippery slope eventually recreating regex. Before a big change, a discussion doesn't seem like a bad idea. There might be an even easier way or additional considerations.

Rational:
1. New edge case 
2. It could help clean up #103 
3. If this was done, all the very expensive and heap allocating clone()s on the Enumerate could be removed.
4. None of this is externally pub, so no other projects would break. I believe it would technically be possible to even have the return not be an owned String but str slices from the original &str too. This would make it so there's no heap allocations, but would also break existing code. Perhaps worth considering before 1.0.0.

Grinkers:
Fixes #105

### DIFF
--- a/src/deserialize/parse.rs
+++ b/src/deserialize/parse.rs
@@ -1472,75 +1472,73 @@ mod test {
     #[test]
     fn parse_exp() {
         let mut edn = "5.01122771367421E15".chars().enumerate();
-        let res = parse(edn.next(), &mut edn).unwrap();
-
-        assert_eq!(res, Edn::Double(5011227713674210f64.into()))
+        assert_eq!(
+            parse(edn.next(), &mut edn),
+            Ok(Edn::Double(5011227713674210f64.into()))
+        );
     }
 
     #[test]
     fn parse_numberic_symbol_with_doube_e() {
         let mut edn = "5011227E71367421E12".chars().enumerate();
-        let res = parse(edn.next(), &mut edn).unwrap();
-
-        assert_eq!(res, Edn::Symbol("5011227E71367421E12".to_string()))
+        assert_eq!(
+            parse(edn.next(), &mut edn),
+            Ok(Edn::Symbol("5011227E71367421E12".to_string()))
+        );
     }
 
     #[test]
     fn parse_exp_plus_sign() {
         let mut edn = "5.01122771367421E+12".chars().enumerate();
-        let res = parse(edn.next(), &mut edn).unwrap();
-
-        assert_eq!(res, Edn::Double(5011227713674.210f64.into()))
+        assert_eq!(
+            parse(edn.next(), &mut edn),
+            Ok(Edn::Double(5011227713674.210f64.into()))
+        );
     }
 
     #[test]
     fn parse_float_e_minus_12() {
         let mut edn = "0.00000000000501122771367421".chars().enumerate();
-        let res = parse(edn.next(), &mut edn).unwrap();
-        assert_eq!(res, Edn::Double(5.01122771367421e-12.into()))
+        assert_eq!(
+            parse(edn.next(), &mut edn),
+            Ok(Edn::Double(5.01122771367421e-12.into()))
+        );
     }
 
     #[test]
     fn parse_exp_minus_sign() {
         let mut edn = "5.01122771367421e-12".chars().enumerate();
-        let res = parse(edn.next(), &mut edn).unwrap();
+        let res = parse(edn.next(), &mut edn);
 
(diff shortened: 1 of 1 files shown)


Grinkers:
@naomijub 

While doing this, I also ran into
https://github.com/Grinkers/edn-rs/blob/symbols/src/deserialize/parse.rs#L69
which I believe should be `(|c| !c.1.is_whitespace() && !DELIMITERS.contains(&c.1))`

when doing tests, I ran into 
```
---- deserialize::test::namespaced_maps_navigation stdout ----
thread 'deserialize::test::namespaced_maps_navigation' panicked at 'assertion failed: `(left == right)`
  left: `Nil`,
 right: `Key(":val")`', src/deserialize/mod.rs:665:9
 
 ---- deserialize::test::namespaced_maps stdout ----
thread 'deserialize::test::namespaced_maps' panicked at 'assertion failed: `(left == right)`
  left: `Key(":abc")`,
 right: `NamespacedMap("abc", Map({"0": Key(":val"), "1": Key(":value")}))`', src/deserialize/mod.rs:572:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I can't find any reference to namespaced maps in EDN (or clojure, for that matter). Is this some sort of reader macro?

```
(clojure.edn/read-string ":abc{ 0 :val 1 :value}")
:abc
```
is what I would have expected and get.

src/deserialize/parse.rs
```diff
@@ -1,7 +1,7 @@
 use crate::edn::{Edn, Error, List, Map, Set, Vector};
 use std::collections::{BTreeMap, BTreeSet};
 
-const DELIMITERS: [char; 5] = [',', ']', '}', ')', ';'];
+const DELIMITERS: [char; 8] = [',', ']', '}', ')', ';', '(', '[', '{'];
```
naomijub, line 4: might be missing set delimiter, but we have a workaround to deal with sets, by replacing the `#{` with `@`

src/deserialize/parse.rs
```diff
@@ -43,10 +43,12 @@ pub fn parse_edn(
     match c {
         Some((_, '\"')) => read_str(chars),
         Some((_, ':')) => read_key_or_nsmap(chars),
-        Some((_, '-')) => Ok(read_number('-', chars)?),
+        Some((_, n)) if n.is_numeric() => Ok(read_number(n, chars)?),
+        Some((_, n)) if (n == '-' || n == '+') && read_number(n, &mut chars.clone()).is_ok() => {
```
naomijub: Maybe this ` read_number(n, &mut chars.clone()).is_ok()` could be a `.peek().next().is_numeric()`?

naomijub: Also, might be better to do  `Some((_, '+') | (_, '-')) if ...` .


Might be `Some((_, '+')) | Some((_, '-')) if ...` , not sure

naomijub:
@Grinkers 
Our implementation of namespace mas is based off https://github.com/edn-format/edn#tagged-elements. It was adapting to Nubank's coding style at the time

naomijub:
@Grinkers I will look this error later today and get back to you

src/deserialize/parse.rs
Grinkers: > 

The 'n' needs to be passed on for later processing of - or + (new edge case mentioned in the edn spec), so can't just be matched and discarded.

Grinkers: > Maybe this ` read_number(n, &mut chars.clone()).is_ok()` could be a `.peek().next().is_numeric()`?

Nice catch.

naomijub:
I'm really sorry for the time we took to review. We were both sick this past month

Issue #106, opened 2023-09-09
Numbers don't match with specification.


Grinkers:
Right now the numbers are 
```
Int(isize),
UInt(usize),
Double(Double),

where:
isize/usize are architecture dependent (hardware pointer size)
Double(i64, u128)
```

https://github.com/edn-format/edn#integers
https://github.com/edn-format/edn#floating-point-numbers

According to the spec, integers should be i64 and floating points should be f64. I don't think we can de arbitrary precision with N or M without software implementations. I think perhaps things should be i64 and f64 in the normal case. With #105 maybe we can match against N to i128? I don't think we can practically match the N/M spec without pulling in dependencies, however the current isize/usize is certainly off-spec compared to the expected types, depending on the target architecture.

Changing this would be a breaking pub API change. Has there been any precedence? Perhaps a 0.18.0?

naomijub:
I don't mind using i64 and u64 for integers, and something like i128/u128 for BigInt. However, there is certainly a problem using f64 for decimal and big decimal as it breaks precision, ordering and hashing. It would require a library like ordered float to wrap that. @evaporei opinions?

evaporei:
I have the same opinion for integers.

As for floating point, I'm ok with bringing ordered float back again into the codebase.

Grinkers:
Is there a reason for a separate u64 on top of i64? I suppose it's not an issue because clojure will read 2^64-1 as a BigInt. When converting from a u64 to a string it should probably add the N character at the end if it is over 2^32. I think this sort of issue is probably more painful than just having an i64 and having the user code have a try_into() (there would likely need an if/match to determine Int vs UInt anyway)

I hadn't considered FP ordering and hashing. I'm not sure clojure handles arbitrary precision here either (cljs almost surely doesn't).

naomijub:
> Is there a reason for a separate u64 on top of i64? I suppose it's not an issue because clojure will read 2^64-1 as a BigInt. When converting from a u64 to a string it should probably add the N character at the end if it is over 2^32. I think this sort of issue is probably more painful than just having an i64 and having the user code have a try_into() (there would likely need an if/match to determine Int vs UInt anyway)
> 
> I hadn't considered FP ordering and hashing. I'm not sure clojure handles arbitrary precision here either (cljs almost surely doesn't).

Basically, I had an issue that the number was larger than i64::MAX. Why not just use all resources available and allow the Rust version to be compatible with larger numbers like i128?

Grinkers:
[...]

PR #110, opened 2023-10-23
Numbers

Grinkers:
WIP for #106 
- [x] pointer size to u64/i64 (breaking change)
- [x] `#[non_exhaustive]` for Edn enum (breaking change)
- [x] investigate/implement ordered-float and f64 in place of Double (almost certainly a breaking change)
- [x] error handle type casting, as right now it does c-style type casting, which will silently ignore bits and give very wrong answers. (behavior change) 
- [x] implement `default-features = false` without sets and no ordered-float dependency

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -30,6 +30,8 @@ jobs:
         run: cargo test --example edn_to_json --features "json"
       - name: example_async
         run: cargo run --example async --features "async"
+      - name: example_no_sets
+        run: cargo run --example struct_from_str --no-default-features
 
   fmt:
       runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,11 +14,14 @@ edition = "2018"
 
 [features]
 async = ["futures"]
+default = ["sets"]
 json = ["regex"]
+sets = ["ordered-float"]
 
 [dependencies]
 regex = { version = "1", optional = true }
 futures = { version = "0.3.5", optional = true }
+ordered-float = { version = "4.1", default-features = false, optional = true }
 
 [dev-dependencies]
 tokio = { version = "1.33", features = ["full"] }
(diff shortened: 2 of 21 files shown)


naomijub:
Do you want us to review as you push or wait til the end?

Grinkers:
Either way.

I'll poke at f64 later this week. I think I'll push BigInt stuff off for later (I don't need it). I'll see how f64 with `ordered-float` goes with an optional `default-features = false` to disable sets. With feature flags, I think it makes sense to either do BigInt correctly or not at all.

naomijub:
Sounds good

Grinkers:
`ordered-float` works with `default-features = false`. The only additional dependency is `num-traits`. 

```
cargo license --avoid-build-deps --avoid-dev-deps --filter-platform x86_64-unknown-linux-gnu
Apache-2.0 OR MIT (1): num-traits
MIT (2): edn-rs, ordered-float
```

I've updated the OP todo list to have sets be optional with no additional dependency and removed BigInt for a future feature.

Grinkers:
Pushed an implementation of `--no-default-features`. It ended up being way more messy than I was expecting.  Two things I really don't like are
1. Test organization. I wanted to take a pass at error handling as a separate issue, probably along with a borrow feature. I'd probably revisit it then
2. Currently `cargo t --no-default-features` fails because it is compiling the docs in src/macros/mod.rs. This could be "fixed" with doctest = false, but not really ideal at all. Thoughts?

https://doc.rust-lang.org/cargo/commands/cargo-test.html for reference

I'll have to study how other crates handle lots of features.

naomijub:
The macro_rules section is something that I particularly feel as less useful now a day. If that Is the issue, I would definitely consider removing that macro

Grinkers:
> The macro_rules section is something that I particularly feel as less useful now a day. If that Is the issue, I would definitely consider removing that macro

I don't mind supporting the macro. It's just the rust in the docs end up getting run as part of tests, but it uses sets. 

Next week I'll be using this branch for my project and I'll see how it goes. I need the better number handling, but don't need sets (and thus no reason to bring in a dependency).

Grinkers:
This got too big. Closing it and breaking things apart. I'll keep the branch alive until everything else gets settled/merged nicely.

PR #113, opened 2023-10-29
64 bit ints instead of pointer size + TryFrom instead of unsafe `as`

Grinkers:
Fixes half of #106

### DIFF
--- a/examples/struct_from_str.rs
+++ b/examples/struct_from_str.rs
@@ -3,7 +3,7 @@ use edn_rs::{Deserialize, Edn, EdnError};
 #[derive(Debug, PartialEq)]
 struct Person {
     name: String,
-    age: u64,
+    age: u8,
 }
 
 impl Deserialize for Person {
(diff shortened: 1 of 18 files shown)


examples/struct_from_str.rs
```diff
@@ -46,19 +46,37 @@ fn person_mistyped() -> Result<(), EdnError> {
     Ok(())
 }
 
+fn person_overflow() -> Result<(), EdnError> {
+    let overflow_edn_str = "  {:name \"rose\" :age 9002  }  ";
```
naomijub, line 50: Doesn't Edn/Clojure ignore whitespaces?

src/edn/mod.rs
```diff
@@ -821,10 +822,14 @@ pub(crate) fn rational_to_double(r: &str) -> Option<f64> {
...
 
 #[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum Error {
     ParseEdn(String),
     Deserialize(String),
     Iter(String),
+    TryFromInt(std::num::TryFromIntError),
+    #[doc(hidden)]
+    Infallable(), // Makes the compiler happy for converting u64 to u64 and i64 to i64
```
naomijub, line 832: Nice!

naomijub:
LGTM, just that one question

examples/struct_from_str.rs
Grinkers: This was just a copy/edit from 
https://github.com/naomijub/edn-rs/blob/554d715adad40cb39b85ee49997b6cf9b40df66d/examples/struct_from_str.rs#L19

This is supposed to fail (changed the `Person` struct to be a u8). It was just the easiest place to add a test because of the `impl Deserialize`, but perhaps it's a realistic example (an error for a person aged over 255).

naomijub (approved):

PR #114, opened 2023-10-29
Double is now f64, leveraging rust's IEEE-754 + optional feature "sets"

Grinkers:
This is a continuation of #110. 

This should be after #113. Not worth dealing with merge conflicts, I'll just rebase to keep the history clean.

### DIFF
--- a/src/json/mod.rs
+++ b/src/json/mod.rs
@@ -11,7 +11,15 @@ pub fn display_as_json(edn: &Edn) -> String {
         Edn::Symbol(s) | Edn::Str(s) => format!("{s:?}"),
         Edn::Int(n) => format!("{n}"),
         Edn::UInt(n) => format!("{n}"),
-        Edn::Double(n) => format!("{n}"),
+        Edn::Double(n) => {
+            // Rust formats an f64 with a value of 2^5 as "32".
+            // We do this to ensure all precision is printed if available, but still adds a decimal point for json.
+            let mut s = format!("{n}");
+            if !s.contains('.') {
+                s.push_str(".0");
+            }
+            s
+        }
         Edn::Rational(r) => format!("{}", rational_to_double(r).unwrap()),
         Edn::Char(c) => format!("'{c}'"),
         Edn::Bool(b) => format!("{b}"),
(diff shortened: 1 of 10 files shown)


Grinkers:
Rebased and pushed, while allowing all tests to pass. I think the implementation is ready for review, but I'm not too happy with how many `#[cfg")]` are scattered about. Some amount is unavoidable, but in particular the tests tests are a little ugly. I took a look at some other crates and the common pattern there seems to be something like

```
// some_tests.rs
mod generic_tests;
#[cfg(feature = "sets")]
mod tests_with_sets;
...
```

I wanted to take a look at Error handling and a borrow impl (less heap allocations, requiring a walker). I think that might be the time/place to fix/clean this up.  
https://github.com/rust-embedded-community/serde-json-core/blob/9deb2b91ea737df79f33f7536587b59ed5f141ee/src/de/mod.rs#L23
for reference for what I sort of had in mind. These can now be added now that we have `non_exhaustive`

src/json/mod.rs
```diff
@@ -11,7 +11,15 @@ pub fn display_as_json(edn: &Edn) -> String {
         Edn::Symbol(s) | Edn::Str(s) => format!("{s:?}"),
         Edn::Int(n) => format!("{n}"),
         Edn::UInt(n) => format!("{n}"),
-        Edn::Double(n) => format!("{n}"),
+        Edn::Double(n) => {
+            // Rust formats an f64 with a value of 2^5 as "32".
+            // We do this to ensure all precision is printed if available, but still adds a decimal point for json.
+            let mut s = format!("{n}");
+            if !s.contains('.') {
```
evaporei, line 18: Maybe I missed it, but is there a test for this?

evaporei (approved):

src/json/mod.rs
Grinkers: Right below, line 136. Actually got a bit lucky because https://github.com/naomijub/edn-rs/blob/master/examples/edn_to_json.rs let me find the edge case. It's a double check, but I thought it should be in this file too, checking `32f64`

evaporei: Nice!

PR #116, opened 2023-11-20
Lints

Grinkers:
Added lints `future-incompatible` and `rust_2018_idioms` and appropriate changes.

https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo
Now using cargo's lint configuration from 1.74. In the near future, the CI can also be changed to

```
- run: cargo clippy --all-features -- --deny warnings
- run: cargo clippy --no-default-features -- --deny warnings
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -55,5 +55,5 @@ jobs:
         - uses: dtolnay/rust-toolchain@stable
           with:
             components: clippy
-        - run: cargo clippy --all-features -- -W clippy::all -W clippy::pedantic -W clippy::nursery --deny warnings
-        - run: cargo clippy --no-default-features -- -W clippy::all -W clippy::pedantic -W clippy::nursery --deny warnings
+        - run: cargo clippy --all-features -- -W future-incompatible -W rust_2018_idioms -W clippy::all -W clippy::pedantic -W clippy::nursery --deny warnings
+        - run: cargo clippy --no-default-features -- -W future-incompatible -W rust_2018_idioms -W clippy::all -W clippy::pedantic -W clippy::nursery --deny warnings
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,13 @@ keywords = ["EDN"]
 license = "MIT"
 edition = "2018"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lints.rust]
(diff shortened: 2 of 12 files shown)


evaporei (approved):

Grinkers:
@naomijub please delete branch https://github.com/naomijub/edn-rs/tree/lints this is the correct/cleaned version

evaporei:
I tried to, but @naomijub has to do it, I don't have access to the repo Settings page :confused:

naomijub:
WOW, I can't do it as well. Let me check if there is a way to delete it 

EDIT:
There was a weird ruleset on deletion

Grinkers:
I think it'd be nice to remove those rulesets on non-master branches (which allow for cleaning up commits, quick mistakes like this missing async, etc). In particular allowing --force. I don't really mind doing my messy --force elsewhere too. The most strict rules are great for master.

Speaking of me missing async @naomijub what was the purpose of the async feature? I keep forgetting to check it because it's not really on my mind. I'm either missing something or maybe it's a legacy thing from pre-2018? I know rust async has gone through a rough road. From my understanding, there's no point in making anything in this library async because serialization and deserialization require all data to be fully `await`ed already. There's no support for incrementally loading of partial bytes as they come in or go out, nor does that really make much sense.

Just as an example, off master right now

Cargo.toml
```
edition = "2021"

[dependencies]
edn-rs = { git = "https://github.com/Grinkers/edn-rs.git", default-features = false }
tokio = { version = "1", features = ["full"] }
```
main.rs
```
use std::str::{self, FromStr};

use edn_rs::{edn, Edn, Vector};
use tokio::fs::File;
use tokio::io::AsyncReadExt;

#[tokio::main]
async fn main() -> std::io::Result<()> {
    let mut file = File::open("foo.txt").await?; // assume the file exists and has valid edn
    let mut contents = vec![];
    file.read_to_end(&mut contents).await?;

    let edn = Edn::from_str(str::from_utf8(&contents).unwrap());
    println!("{edn:?}");

    let edn = edn!([1 1.5 "hello" :key]);
    println!("{edn:?}");
    Ok(())
}
```
fully works as expected, without the feature.

naomijub:
I removed the delete rule.

Async feature was created before async await was fully stable, so I guess now they are useless

Grinkers:
I will clean it up, depending on what we do with #117

PR #118, opened 2023-11-24
Bump to rust edition 2021 + best practices.

Grinkers:
The `async` feature has been removed.
The async example is now using stable async/futures.

Relevant discussions:
https://github.com/naomijub/edn-rs/pull/116
https://github.com/naomijub/edn-rs/issues/117

I'm not sure how many people use this crate, but I think this might be one of the last big breaking changes for standard features. A `0.18.0` release might be a good idea. An added note about an org change (https://github.com/naomijub/edn-rs/issues/109) from `0.19.0` might be a good idea.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,8 +40,8 @@ jobs:
           name: Run example tests with json feature
           command: cargo test --example json_to_edn --features "json"
       - run:
-          name: Run example tests with async feature
-          command: cargo run --example async --features "async"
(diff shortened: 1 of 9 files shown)


naomijub (approved):
I will bump to 0.18
We should decide the org name

Grinkers:
@naomijub hold that release for https://github.com/naomijub/edn-rs/issues/119
hopefully the last breaking change I suggest

PR #120, opened 2023-11-24
EdnError 1.0.0 (std) pub API

Grinkers:
Draft so we can bikeshed https://github.com/naomijub/edn-rs/issues/119

PR #121, opened 2023-12-08
API reduction

Grinkers:
Discussion bits are scattered about, but mostly in #120 (which was growing beyond scope).


Grinkers:
@naomijub I'm able to merge into master without a review. That's probably not what we want. Is it possible to add all the restrictions on master but not other branches?

naomijub:
Will do, I probably can review it after my lunch and then add ruleset to main

Grinkers:
I sneaked in test refactoring here because it was causing noise for #120. 

I'm guessing for historical reasons it made it easier to work with, but it was testing internal structures that I intend to rewrite, so I converted them to integration tests in https://github.com/naomijub/edn-rs/pull/121/commits/78bdf1cbc201b0295a44c04ff6633e73a6b827bb

If it's ok, I'll continue doing this as I run into things like in https://github.com/naomijub/edn-rs/pull/121/commits/6db3875a4b4afe1ecc2588ff328c301ad2c6c396. A lot of the tests are actually using only public interfaces so can live in `tests/`. As you probably noticed, there's a lot less `#[cfg(feature = "sets")]` scattered all over too. 

I think going forward cut/paste. I'll keep them in separate commits so it's easy to track.

naomijub:
> I sneaked in test refactoring here because it was causing noise for #120.
> 
> I'm guessing for historical reasons it made it easier to work with, but it was testing internal structures that I intend to rewrite, so I converted them to integration tests in [78bdf1c](https://github.com/naomijub/edn-rs/commit/78bdf1cbc201b0295a44c04ff6633e73a6b827bb)
> 
> If it's ok, I'll continue doing this as I run into things like in [6db3875](https://github.com/naomijub/edn-rs/commit/6db3875a4b4afe1ecc2588ff328c301ad2c6c396). A lot of the tests are actually using only public interfaces so can live in `tests/`. As you probably noticed, there's a lot less `#[cfg(feature = "sets")]` scattered all over too.
> 
> I think going forward cut/paste. I'll keep them in separate commits so it's easy to track.

I'll probably take a couple of days to review the changes and test them

src/deserialize/mod.rs
```diff
@@ -445,345 +409,4 @@ pub fn from_edn<T: Deserialize>(edn: &Edn) -> Result<T, Error> {
...
-            ))
-        );
-    }
-
-    #[test]
-    fn string_with_empty_set() {
-        assert_eq!("\"#{}\"", format!("{}", Edn::from_str("\"#{}\"").unwrap()));
-    }
-}
+mod test {}
```
naomijub: I think this can be removed if we have nothing there

src/edn/mod.rs
```diff
@@ -712,10 +709,7 @@ impl std::str::FromStr for Edn {
 
     /// Parses a `&str` that contains an Edn into `Result<Edn, EdnError>`
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let clean = String::from(s.trim_start());
```
naomijub, line 715: this had a nice reason to exist, it broke when there was whitespaces before the edn

Grinkers: I want to remove it so the walker impl can give accurate feedback.
`"                   {"` should throw a nice error at line 1 column 20, not line 1 column 2

src/deserialize/mod.rs
Grinkers: I can remove this one. I'll keep the one in `deserialize/parse.rs` as I'm sure there will be unit tests with the walker impl

src/edn/mod.rs
naomijub: that makes sense, one thing that has always annoyed me is the fact that this ` edn.chars().enumerate()` doesnt preserve the initial enumeration, but associating the enumerate number to the iterator, means we actually need to consume it in the first pass. And that is slow

Grinkers: I actually think the walker impl won't create/allocate an owned string at all. I was mostly just trying to clean things up to move to a 1.0 api (both public and private).

naomijub (approved):
LGTM
Nice work

Issue #119, opened 2023-11-24
EdnError relies on an owned and allocated String to convey information


Grinkers:
This is problematic for a few reasons. 
1. This makes it very hard for non-std or to do TDD without having to change everything
2. Very hard to programmatically handle different edge cases 

A couple examples
https://github.com/naomijub/edn-rs/blob/d852751abfa90af4e4ed57ba9d807137d80802a6/src/deserialize/parse.rs#L1114
https://github.com/naomijub/edn-rs/blob/d852751abfa90af4e4ed57ba9d807137d80802a6/src/deserialize/parse.rs#L648

Here's some references for serde json
https://docs.rs/serde_json/latest/src/serde_json/error.rs.html#17-22
embedded
https://github.com/rust-embedded-community/serde-json-core/blob/9deb2b91ea737df79f33f7536587b59ed5f141ee/src/de/mod.rs#L23

I quite like serde_json's process of being very abstract https://docs.rs/serde_json/latest/serde_json/error/struct.Error.html
Essentially I think the end use case is 99% of the time you handle the case as "there was an error" and log it. in the 1% case, you have functions to describe the error for the end-user. This abstract behavior will also make it easy to write tests, make internal changes without API changes, and make other environments easier (no-std or heapless).

I will be taking a pass on std performance and I consider this a prerequisite. As far as a 1.0.0 interface, I think this the last breaking change needed. a `heapless` or `no-std` feature doesn't currently exist, so it can't be breaking. I don't think there's much point in supporting a `no-std`, but I don't want to eliminate the possibility. I will probably take an attempt for heapless for constant time performance and embedded, but I don't consider this a needed thing for 1.0.0 (a reasonable 1.1.0).

I think this will end up being a pretty big rework and series of commits. The biggest performance issue we have now is all the `clone()`s, which requires a new heap allocation, `memcpy`, get a count or position, and then we throw it all away. I will need to implement a walker of sorts, so a stable target would be of great help.

More people the better for feedback on the api/structure.

Grinkers:
https://github.com/Grinkers/edn-rs/commit/e2e8f5f9778948fa8160fedafcae0e06470dcf14

A quick proof of concept. Very much not ready. Tests are all sorts of messed up because of a lack of Eq/PartialEq. Examples `edn_from_str` and `async` at the very least work without changes, just to show that it's not *that* breaking when in the 99% case.

Grinkers:
@naomijub @evaporei 
[...]

PR #123, opened 2023-12-10
EdnError rework

Grinkers:
Continuation of #120 with a rebase (should be a lot less noise). Currently a draft for #119

### DIFF
--- a/src/edn/error.rs
+++ b/src/edn/error.rs
@@ -1,0 +1,93 @@
+use alloc::boxed::Box;
+use alloc::string::{String, ToString};
+use core::fmt::{self, Debug, Display};
+
+pub struct Error {
+    code: Code,
+    // line: usize,   // TODO walker
+    // column: usize, // TODO walker
+}
+
+#[non_exhaustive]
+pub enum Code {
+    /// Catchall/placeholder error messages
+    Message(Box<str>),
+    Io(std::io::Error),
+    TryFromInt(std::num::TryFromIntError),
+    #[doc(hidden)]
+    Infallable(), // Makes the compiler happy for converting u64 to u64 and i64 to i64
+}
+
+impl Error {
+    pub(crate) fn ParseEdn(owned_string: String) -> Self {
+        // TODO remove, just so things can compile with less changes
+        Self {
+            code: Code::Message(owned_string.into_boxed_str()),
+        }
+    }
+    pub(crate) fn Deserialize(owned_string: String) -> Self {
+        // TODO remove, just so things can compile with less changes
+        Self {
+            code: Code::Message(owned_string.into_boxed_str()),
+        }
+    }
+    pub(crate) fn Iter(owned_string: String) -> Self {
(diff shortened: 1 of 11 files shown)


Grinkers:
Oops, missed a couple tests that weren't run on my local setup. Fixed.

Please look/think about the new Error type and new usage. As it is right now, only `Display` and `Debug` are public. `Eq` and `PartialEq` are _not_.  I'll probably make the error `code` pub too. I'll have to see how the walker impl ends up.

Obviously the TODOs are to do.

src/edn/error.rs
```diff
@@ -0,0 +1,91 @@
...
+pub enum Code {
+    /// Catchall/placeholder error messages
+    Message(Box<str>),
+    Io(std::io::Error),
+    TryFromInt(std::num::TryFromIntError),
+    #[doc(hidden)]
+    Infallable(), // Makes the compiler happy for converting u64 to u64 and i64 to i64
+}
+
+impl Error {
```
naomijub, line 21: I think this functions should be snake_cased

Grinkers: These functions aren't intended to remain in the code base. It's only a super hacky way to make things 
like `return Err(Error::ParseEdn(...))` compile without changing the code. 

I put TODOs in the comments, but a walker wouldn't be using these at all.

Grinkers:
btw I started fuzzing with
https://github.com/Grinkers/edn-rs_fuzzer/tree/main
finding lots of things. Errors are also pretty annoying because I'll sometimes just end up with "ParseEdn" (not sure why on this one) or something at "index 2" but it's actually 500 characters deep.

Grinkers:
Any other thoughts before doing the implementation?

I'd like to see #122 and #125 finished before progressing, as there's already conflicts.

Grinkers:
While looking at clojure's read-string limitations for symbols I ran into 
https://github.com/naomijub/edn-rs/blob/bd558559f2b36c7c1c2ba3d24b38c15aaf0fcab5/tests/deserialize.rs#L803

```
(clojure.edn/read-string "5011227E71367421e12")
java.lang.NumberFormatException: Invalid number: 5011227E71367421e12 
```

I'll be expanding my fuzzer for things that are not valid edn. Here's a couple patterns that it has found that we consider valid EDN but is not.
```
"#'clojure.generated.ns1/var1"
"#object[clojure.lang.Namespace 0x6dcc40f5 \\\"clojure.generated.ns1\\\"]"
```

I don't think these issues should block #125, as that's fixing valid edn issues. I think I'll rebase after 125, implement a walker with the most minimal version of EdnError as drafted here (while making sure there's no regressions), then start fuzzing non-valid edn too.

For the new ErrorCode enum, I think I'll loosely try to match the names Clojure's reader gives. https://github.com/clojure/clojure/blob/c6756a8bab137128c8119add29a25b0a88509900/src/jvm/clojure/lang/EdnReader.java

PR #124, opened 2023-12-10
`make tests` and `make examples`, combined with `make test`, match CI runners.

Grinkers:
For convenience. I keep forgetting what to run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,18 @@ examples:
 	cargo test --example json_to_edn --features "json"
 	cargo test --example edn_to_json --features "json"
 	cargo run --example async
+	cargo run --example struct_from_str --no-default-features
 
-doc-tests:
+tests:
+	cargo test --all-features --no-fail-fast
+	cargo test --no-default-features --no-fail-fast
 	cargo test --doc
 
-test: unit integration examples doc-tests
+clippy:
+	cargo clippy --all-features -- -W future-incompatible -W rust_2018_idioms -W clippy::all -W clippy::pedantic -W clippy::nursery --deny warnings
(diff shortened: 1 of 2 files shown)


Grinkers:
@naomijub I can still merge without review

Grinkers:
Added clippy and fmt too...

Makefile
```diff
@@ -10,8 +10,18 @@ examples:
...
+	cargo run --example struct_from_str --no-default-features
 
-doc-tests:
+tests:
+	cargo test --all-features --no-fail-fast
+	cargo test --no-default-features --no-fail-fast
 	cargo test --doc
 
-test: unit integration examples doc-tests
+clippy:
```
naomijub, line 20: With rust 1.74 we can probably move this configurations to something like this:
https://github.com/rewin123/space_editor/blob/main/Cargo.toml#L66

WDYT?

Grinkers: https://github.com/naomijub/edn-rs/blob/master/Cargo.toml#L13
We actually already have this. This is just an explicit check to match CI. Once CI runners have been updated and  we're at rust 1.76 stable we can clean both up.

Grinkers:
I figure the Makefile was mostly just for convenience, right? Seeing as how likely any maintainer would also be into Clojure... Anything against [babashka tasks](https://book.babashka.org/#tasks)?

It's quite convenient. You get a specific error like `Error while executing task: test_lib_features`, without stuff flying by your screen, while doing `bb test`. I also just tested it on windows (no msys, WSL, etc. just downloaded bb.exe with standard windows and standand cmd).

I also added an empty `.rustfmt.toml`. While testing bb tasks on windows, rustfmt was grabbing configs from a parent dir.

PR #125, opened 2023-12-11
Various fixes for problems found while fuzzing valid EDN.

Grinkers:
The two big problems found were not handling characters correctly and failing to parse single digit numbers(!).

I've also implemented Arbitrary Numbers, but handled as strings. This way no data is ever lost, but this goes beyond the scope of std-only. In the future a feature could be added to change the type (or introduce a breaking change if rust std ever gets these features).

Display now more closely matches clojure too.

I've run a few million iterations with my fuzzer https://github.com/Grinkers/edn-rs_fuzzer (using clojure's EDN generator). I'll fuzz incorrect EDN after we decide on the new Error and I implement a walker.


Grinkers:
Sorry for all the force pushes, it's a bit annoying not being able to easily see the results locally. Make was just blowing past everything, so maybe need to come up with something better. I also confused myself as this is on my fork. I'll keep this PR as-is.

Grinkers:
@naomijub 
https://github.com/naomijub/edn-rs/pull/125/commits/7969aad52dba7dea99529da5d430e6112714c80b
is pretty critical. 

Still need feedback on Arbitrary length things. Roundtrip compatible but otherwise not handled (future functions can be added to parse/convert), but the internal structure (String) is probably ok for now

naomijub:
> @naomijub 
> 
> https://github.com/naomijub/edn-rs/pull/125/commits/7969aad52dba7dea99529da5d430e6112714c80b
> 
> is pretty critical. 
> 
> 
> 
> Still need feedback on Arbitrary length things. Roundtrip compatible but otherwise not handled (future functions can be added to parse/convert), but the internal structure (String) is probably ok for now

I will try to review it this weekend

Grinkers:
Conflicts from #122 are resolved with a rebase. No other changes intended.

src/deserialize/parse.rs
```diff
@@ -271,23 +266,52 @@ fn read_number(
...
-        .next()
-        .ok_or_else(|| Error::ParseEdn("Could not identify symbol index".to_string()))?
-        .0;
-    let c = chars.next();
-    c.ok_or(format!("{c:?} could not be parsed at char count {i}"))
-        .map(|c| c.1)
-        .map(Edn::Char)
-        .map_err(Error::ParseEdn)
+        .enumerate()
+        .take_while(|&(i, c)| i <= 200 && !c.1.is_whitespace())
```
naomijub: Theoretically Clojure supports integers with infinite (memory bounded) size right? Maybe we could find a way to improve/increase this now that we are using string?

src/edn/mod.rs
```diff
@@ -165,69 +167,72 @@ impl Map {
...
-                .iter()
-                .map(ToString::to_string)
-                .fold(String::new(), |mut acc, i| {
-                    acc.push_str(&i);
-                    acc.push_str(", ");
-                    acc
-                })
-        )
+        write!(f, "[")?;
+        let mut it = self.0.iter().peekable();
```
naomijub, line 175: Have you checked the impact of making everything in the library peekable? Might be a good idea

naomijub:
I'm actually good with the PR, just take a look at the idea of wrapping everything into a peekable. I would love if @evaporei could code review as well, as this is a larger PR

src/edn/mod.rs
evaporei: I would guess this is faster because there's no need to allocate a new string.

tests/emit.rs
```diff
@@ -0,0 +1,29 @@
...
+    use edn_rs::Edn;
+
+    fn edn_to_string_unwrap(s: &str) -> String {
+        Edn::from_str(s).unwrap().to_string()
+    }
+
+    #[test]
+    fn map_formatting() {
+        let edn = edn_to_string_unwrap("{nil \"`HQ5|>^>rQNL9E..y#}U63/S_Qo- IMpw>gnM']jD\"}");
+        assert_eq!(edn, "{nil \"`HQ5|>^>rQNL9E..y#}U63/S_Qo- IMpw>gnM']jD\"}");
```
evaporei, line 14: :heart:

src/deserialize/parse.rs
```diff
@@ -278,16 +278,43 @@ fn read_number(
...
+        .map(|(_, c)| c.1)
+        .collect::<String>();
+
+    let mut consume_chars = |n| {
+        // We need to map/collect to consume out of the Enumerate
+        let _ = chars.take(n).map(|c| c.1).collect::<String>();
+    };
+
+    match element {
```
evaporei, line 287: I'm sorry, I don't get this part, perhaps my Clojure skills are a bit rusty :drum: 

If I recall correctly, using `\a` is just like `'a'` (char) in Rust. Then, what do those `\\xyz` mean?

I was trying them (`\\newline`, `\\space`, etc) in the [REPL](https://tryclojure.org/), and only `\\newline` resolved to something.

README.md
```diff
@@ -368,6 +368,7 @@ fn complex_ok() -> Result<(), EdnError> {
     - [x] nil `""`
     - [x] String `"\"string\""`
     - [x] Numbers `"324352"`, `"3442.234"`, `"3/4"`
+    - [x] Arbitrary Numbers `"4650481213195981094N"`, `"-6.731982E-96M"`. These are stored as strings and not checked for correctness
```
evaporei: :tada:

Grinkers:
Bump. It'd be really good to get this reviewed.
https://github.com/edn-rs/edn-rs/pull/125/commits/1942a7d81f3a07db115e5161e18a5af93c8ecb58
is a very critical bug fix.

I can break out the Arbitrary numbers if it that's the holdup.

src/deserialize/parse.rs
naomijub: @Grinkers ?

src/edn/mod.rs
naomijub: @Grinkers ?

Grinkers:
Rebased against main.

I removed the arbitrary length int/floats. I'm not happy with it long-term, as I don't feel it's suitable for a 1.0 target. I moved that to https://github.com/edn-rs/edn-rs/pull/134 as a draft while I think about it. This PR as-is should be considered high priority due to not following the spec.

Grinkers:
Rebased/conflicts resolved, again.

Grinkers:
Would it help if I broke things up into more pieces so it's easier to review and track?
@naomijub @evaporei

Grinkers:
Rebased in https://github.com/Grinkers/edn-rs/pull/3 and part of #140

PR #127, opened 2023-12-15
Fix bb script (clippy isn't test)

Grinkers:


Grinkers:
@naomijub I can still rebase and merge without review. I can't when CI doesn't pass, like in the EdnError draft.

naomijub:
Can you check if the new ruleset works?

Grinkers:
> Can you check if the new ruleset works?

Looks good. I can't merge #125 without a review.

PR #130, opened 2023-12-18
Fix serialize/deserialize implementation for edn-derive

evaporei:
This PR fixes `edn-derive` (currently failing the tests).

The commit messages explain the fix :slightly_smiling_face: 

Related to https://github.com/edn-rs/edn-derive/issues/32

### DIFF
--- a/src/edn/utils/index.rs
+++ b/src/edn/utils/index.rs
@@ -20,6 +20,7 @@ impl Index for u64 {
             return match *v {
                 Edn::Vector(ref vec) => vec.0.get(idx),
                 Edn::List(ref vec) => vec.0.get(idx),
+                Edn::Map(ref map) => map.0.get(&self.to_string()),
                 _ => None,
             };
         }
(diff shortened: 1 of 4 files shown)


naomijub:
Does this solve the issue?

evaporei:
Almost, [this test](https://github.com/edn-rs/edn-derive/blob/master/tests/unnamed_struct_deserialize.rs) still fails, but I _think_ we can remove it, can you answer [this question](https://github.com/edn-rs/edn-derive/issues/32#issuecomment-1861598138) so I can decide what to do with it?

naomijub:
> Almost, [this test](https://github.com/edn-rs/edn-derive/blob/master/tests/unnamed_struct_deserialize.rs) still fails, but I _think_ we can remove it, can you answer [this question](https://github.com/edn-rs/edn-derive/issues/32#issuecomment-1861598138) so I can decide what to do with it?

Done

src/edn/utils/index.rs
```diff
@@ -17,6 +17,7 @@ impl Index for u64 {
             return match *v {
```
Grinkers, line 20: Right above here, I believe this should actually be 
`impl Index for usize`? not `impl Index for u64` as this is a dereference via pointer. It would be nice to have this fixed for 32 bit systems so they don't need a double width pointer just to index

Grinkers: I don't think this would ever result in a crash or problem, but it'd be nice to clean up my mess in the same PR. Thanks!

Grinkers (approved):

PR #131, opened 2023-12-19
The crate is now no_std with no breaking changes with default-features. 

Grinkers:
This ended up being a lot easier that I expected.

WIP example compiles with
`cargo build --release --target thumbv6m-none-eabi`

I'll make some better examples and probably make it a separate project. Things get really weird with the current examples and dev-dependencies, not to mention we can't run it anyway. I'll make sure to grab some dev boards from the office so I can make some real examples over the holidays. I'll probably write a tiny little lisp that controls some hardware via s-expressions over serial.


Grinkers:
Closing for non-fork PR.

PR #134, opened 2023-12-21
WIP Implement Arbitrary int and floats

Grinkers:
`ArbitraryInt` and `ArbitraryFloat` are pub and I don't feel this is good target we want long-term for a 1.0 release. It makes no sense for a user to be able to do something like
`let invalid_edn = Edn::ArbitraryInt("lolfoobar");`

On that note @naomijub what was the thought of having Rational be a String? I think this has the same problem. It should probably be a tuple of `(i64, u64)`? Interestingly I can't find any references to this in the EDN spec.


Grinkers:
https://github.com/nadavrot/arpfloat
Candidate library for proper Arbitrary types. No additional dependencies and also `no_std` for embedded. The thought is to add another additive feature.

naomijub:
> `ArbitraryInt` and `ArbitraryFloat` are pub and I don't feel this is good target we want long-term for a 1.0 release. It makes no sense for a user to be able to do something like `let invalid_edn = Edn::ArbitraryInt("lolfoobar");`
> 
> On that note @naomijub what was the thought of having Rational be a String? I think this has the same problem. It should probably be a tuple of `(i64, u64)`? Interestingly I can't find any references to this in the EDN spec.

I don't remember the reason, @evaporei do you remember?

naomijub:
> https://github.com/nadavrot/arpfloat Candidate library for proper Arbitrary types. No additional dependencies and also `no_std` for embedded. The thought is to add another additive feature.

This crate seems solid, I didn't know it

PR #135, opened 2023-12-22
More aggressive CI

Grinkers:
Run CI on Windows and embedded. More Linux targets. All using stable minus 2 releases.
Run CI on all branches, not just when doing a PR.

Rebased and cleaned up conflicts from #132.

Tests and Builds are separated. 
We test both 32 and 64 bit on Windows and Linux.
We build for all major architectures, with and without no_std (but can't execute/run tests). Also one embedded target, Cortex-M0

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -2,46 +2,154 @@ name: Rust
 
 on:
   push:
-    branches: [ "main" ]
+    branches: [ "*" ]
   pull_request:
     branches: [ "*" ]
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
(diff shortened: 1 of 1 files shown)


Grinkers:
In a couple stable releases, we can remove `clippy_check` in favor of just the Cargo.toml lint configurations.

Grinkers:
Rebased in https://github.com/Grinkers/edn-rs/pull/2 and part of #140

Grinkers:
@evaporei re-opened to start unraveling
https://github.com/edn-rs/edn-rs/pull/135

naomijub (approved):

Issue #139, opened 2023-12-24
Upstream (merge/rebase) continued progress


Grinkers:
@naomijub is on vacation https://github.com/edn-rs/edn-derive/issues/34 

It is becoming impossible to keep every PR and issue completely isolated without conflicts.  Any new changes are indefinitely being blocked by other blocked PRs. It's becoming harder to manage the conflicts and isolation than the code.

I'm going to continue progress on https://github.com/Grinkers/edn-rs (main branch). I'll try to keep things clean (main has no force/rewriting history) with a public PR with a description, justification, and/or notes for each category of issue. The intention is to eventually upstream changes. We'll see where things are in Feb.

PR #140, opened 2023-12-24
`edn-rs/edn-rs` <- `Grinkers/edn-rs`

Grinkers:
Issue discussion: 
https://github.com/edn-rs/edn-rs/issues/139

### DIFF
--- a/src/edn/mod.rs
+++ b/src/edn/mod.rs
@@ -230,6 +230,16 @@ impl core::fmt::Display for Map {
     }
 }
 
+fn char_to_edn(c: char) -> String {
+    match c {
+        '\n' => "\\newline".to_string(),
+        '\r' => "\\return".to_string(),
+        ' ' => "\\space".to_string(),
+        '\t' => "\\tab".to_string(),
+        _ => format!("\\{c}"),
+    }
+}
+
 impl core::fmt::Display for Edn {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         let text = match self {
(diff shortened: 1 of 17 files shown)


Grinkers:
https://github.com/Grinkers/edn-rs/pull/2

Grinkers:
edn-rs now passes roundtrip with clojure (jvm)'s own edn fuzzer
https://github.com/Grinkers/edn-rs/pull/3

Grinkers:
Did a --force rebase with https://github.com/edn-rs/edn-rs/pull/130 (just to keep things clean).

Grinkers:
trait Serialize uses reference to self instead of move
https://github.com/Grinkers/edn-rs/pull/4

Grinkers:
Edn::Rational from String to (i64, u64)
https://github.com/Grinkers/edn-rs/pull/5

Grinkers:
EdnError rework with parse rewrite
https://github.com/Grinkers/edn-rs/pull/6

src/edn/mod.rs
```diff
@@ -536,7 +540,7 @@ impl Edn {
     /// use edn_rs::edn::{Edn, Vector};
     ///
     /// let edn = Edn::Vector(Vector::new(vec![Edn::Int(5), Edn::Int(6), Edn::Int(7)]));
-    /// let expected = "[5, 6, 7, ]";
+    /// let expected = "[5 6 7]";
```
naomijub, line 543: Both are valid right?

src/edn/mod.rs
```diff
@@ -780,15 +719,6 @@ mod test {
     use alloc::vec;
 
     use super::*;
-    #[test]
```
naomijub: do we have these tests covered elsewhere?

Grinkers: I removed them because with `(i64, u64)` they're infallable

Grinkers:
@naomijub Thanks for starting to look at this. Instead of dealing with this huge draft PR, with some details inside that we may not want to upstream without discussion, I'll start breaking them apart piece by piece and squashing as we go. I mostly put this here for me to track.

We'll have to do it linearly though, I don't want to rebase/isolate 15 separate times; it's work built on top of other work. For example, Rational can be looked at in isolation, but after some other things.

Grinkers:
Thinking about it, I'll start rebasing against edn-rs/main and deleting comments as we go. Think of this as more of an automatic linear issue tracker. Will delete when there's nothing left!

PR #142, opened 2024-01-03
Display is now closer behavior to clojure's pr-str of edn. 

Grinkers:
Mostly cosmetic, as having extra `,` and such is within spec.

I think before there was concern about `iter().peekable()`? Or maybe misunderstandings from a bunch of other commits. Here's a small example of this pattern, but it gets completely optimized out. This is actually faster because there's no String being allocated, and it's going straight into the fmt buffer.
https://godbolt.org/z/n1qfbTWs8


Grinkers:
Probably easier to view the two commits in isolation. One is just a rename, but makes the text diff large.

evaporei (approved):

Issue #136, opened 2023-12-22
1.0.0 pub API


Grinkers:
Not including #119 what's the best way to discuss and coordinate all the breaking changes/behaviors to get to 1.0.0? I think if we try to do it all in one go in one place, nobody is going to be able to follow or track all the discussions/bikeshedding. I'll do the work, but would like to be efficient about it.

Also, from a high level view, what's the priority? 
Do we want things to be more like clojure or more like rust? Given it's EDN, and probably almost always going to be used with clojure(script), maybe it should be more like clojure?

Here's a couple examples of things I'm messing around with on https://github.com/Grinkers/edn-rs/tree/bikeshed

clojure
```clojure
(boolean "foobar") => true
(int "123") => java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Character [at <repl>:2:1]
```

edn-rs
```rust
Edn::Str("foobar".to_string()).to_bool() -> None
Edn::Str("123".to_string()).to_int() -> Some(123)
```

Grinkers:
Closing in favor of #139 

I will try to keep objective things separate from subjective things, so subjective things can be discussed and changed more easily

Grinkers:
```rust
Edn::Int(42).to_char() -> None
``` 

```clojure
(char 42) -> \*
```

I think a lot of the methods here https://docs.rs/edn-rs/latest/edn_rs/edn/enum.Edn.html#implementations are confusing. I'm thinking we should remove it all or specify a spec (eg. try to match `clojure.core`). I really wish there was an easy way to iterate over all collections, like clojure. I think it can only be done with a new type.

Grinkers:
https://github.com/edn-rs/edn-rs/issues/158

PR #153, opened 2024-05-01
`to` collections return reference instead of move data out of `Edn`.

Grinkers:
This is a breaking change, but right now iterating over nested collections is a huge pain. It requires cloning the entire Edn object each time.

I'm not particularly happy with the naming, but it's the least breaking version for now. Another example of needing to think really hard on https://github.com/edn-rs/edn-rs/issues/136. What we really want is a something like [seq](https://clojuredocs.org/clojure.core/seq) but I'm not sure how viable that is in rust.

### DIFF
--- a/src/edn/mod.rs
+++ b/src/edn/mod.rs
@@ -102,8 +102,8 @@ impl Vector {
     }
 
     #[must_use]
-    pub fn to_vec(self) -> Vec<Edn> {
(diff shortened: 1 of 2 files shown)


naomijub (approved):
Removing those clones was always part of our plan, but I guess we forgot

Thanks

PR #155, opened 2024-05-29
Remove empty

Grinkers:
```clojure
(clojure.edn/read-string "")
nil
(clojure.edn/read-string " ")
nil
(clojure.edn/read-string " \n  ; foobar")
nil
```
There is no Empty in the EDN spec.

This PR currently includes #150. It is impractical to backport all changes.
@naomijub I can add this to #150 or keep it here in a draft.


Grinkers:
https://github.com/edn-rs/edn-rs/issues/158

PR #161, opened 2025-05-22
New lelwel parser

naomijub:


naomijub:
Less efficient than original parser

